PR #334, opened 2020-04-25
Correct link to NUnit 3 GUI

mikkelbu:


mikkelbu:
No problem :smile: I was just looking into placing the analyzer documentation here.

Issue #352, opened 2020-06-08
Docfx: Build Process -- run html-proofer or similar to check for dead links


SeanKilleen:

SeanKilleen:
#476 wasn't successful for html-proofer due to it hitting github and getting rate limited. I'd like us to be better internet citizens.

I think we can follow-up on this with another PR and perhaps ignore github issue and PR URLs. That should cut down on a ton of noise.

For what it's worth, here's the GitHub action job that worked:

```yaml
html_proofer:
    name: HTML checks (FYI only)
    needs: [build]
    runs-on: ubuntu-latest
    steps:
    - name: Download site artifacts
      uses: actions/download-artifact@v2
      with:
        name: siteArtifact
    - name: unzip site contents
      run: unzip _site.zip
    - name: Check HTML
      uses: chabad360/htmlproofer@master
      with:
        directory: "./docs/_site/articles"
        arguments: --log-level debug
```

SeanKilleen:
I'm thinking that it probably makes sense to leave this for now. We fixed a bunch of links and I'm OK doing this as a periodic batch process rather than trying to get it into the build process, given the state of the ecosystem.

PR #372, opened 2020-06-13
Use https for links where possible

ChrisMaddock:
Using an old tool I have from a different project, which converts any links to https where possible. 

For ease of review, it ignores anything which redirects, hence why there's still some VS/Google Code links left as http.


SeanKilleen:
Nice! I was going to do this as part of #352 but happy to not have to :)

PR #377, opened 2020-06-13
Docfx build changes

SeanKilleen:


SeanKilleen:
The docfx build is currently doing what it needs to do -- surfaced some good issues around case sensitivity on linux vs. windows. 

Given that, going to merge this and begin addressing some of the warnings in another PR.

PR #380, opened 2020-06-14
Fix case-sensitive links

### LINKED ISSUE #378: Docfx build: Resolve case-sensitivity warnings

SeanKilleen:
Resolves #378.

After this point, there should be no warnings in the docfx build.


SeanKilleen:
That's a nice sight! 

> image

Issue #342, opened 2020-06-08
Docfx: Build / Push to GitHub Pages


SeanKilleen:
Preferably GitHub actions so we can keep it all here and readily visible.

* Build using Docfx
* Push to a `gh-pages` branch.

SeanKilleen:
The docs are now published to the `gh-pages` branch which is then served by GitHub pages.

The site currently lives at http://nunit.org/docs.

I'll be following up quickly with a robots.txt file to delay search engine crawling for a little bit.

I wasn't aware that creating a `gh-pages` branch would publish the site and that I wouldn't be able to simply disable github pages functionality. My best guess is it's because GitHub pages is enabled for the org with a CNAME record so gh-pages in projects within the org are automatically served?

Either way, I can't disable it and I don't think it's necessary to. The site will be live but not crawled once the follow up PR is complete.

/cc @CharliePoole @ChrisMaddock.

ChrisMaddock:
Hey Sean - great news that it’s up and running! I get a 404 on that link, however?

ChrisMaddock:
Ah, never mind, just seen #403. 😊

PR #396, opened 2020-06-17
Add simple publish step with conditional

SeanKilleen:
No actual code yet; just an echo statement.

Supports #342.

Now we're getting to the fun part! This adds a step to the github actions that should only run for the master branch and not PRs. 

To test it, we're making the step an echo statement for now. It should not run on this PR, and then should run once merged.


SeanKilleen:
Indeed, the step was skipped. 👍 

> image

Merging this to see it run against master. This way the next PR can set up the publishing knowing the conditional should work.

PR #408, opened 2020-06-18
Enable the full text search index

### LINKED ISSUE #407: Add Full-text search to generated site
Just realized we weren't taking advantage of this! 

Docfx will supply a search bar. All we have to do is extract the metadata.

SeanKilleen:
Resolves #407.

> full-text-search


ChrisMaddock:
Nice feature! 😄

Issue #350, opened 2020-06-08
Docfx: Ensure TOC is searchable on its own (not just actual pages underneath)


SeanKilleen:
Otherwise, we'll need to make each item in the hierarchy its own landing page as well. People need to be able to search the parent links, even when those links don't directly have content, because those are often the topics that people are seeking.

ChrisMaddock:
Something I've just found - not all our pages are in the TOC, some are just linked from other pages. It would be useful to be able to search for anything by page title though - is there a setting to do that? Or would the solution instead be to add everything to the TOC?

My use-case, I was looking for NUnit-Project-XML-Format, which is only linked from the XML formats page.

SeanKilleen:
@ChrisMaddock I'll look into that as part of this ticket. I just saw something similar with the analyzers a little while ago.

On one hand, maybe all the contents should be in the TOC as subitems, since we can use an xref to refer to any doc in any folder. 

But yeah I'll definitely be looking into it.

ChrisMaddock:
Great, thanks!

SeanKilleen:
Created https://github.com/dotnet/docfx/issues/6074 to hopefully get some more information on this.

SeanKilleen:
Posted the question to SO as well just in case someone has the experience there: https://stackoverflow.com/questions/62421615/how-can-i-surface-toc-items-within-docfx-that-are-not-directly-tied-to-content

SeanKilleen:
OK, I think the key to this might be to use `topicHref`, which I didn't fully understand before.

https://dotnet.github.io/docfx/tutorial/intro_toc.html#link-to-another-toc-file

I'll look into this when i get the chance.

SeanKilleen:
Updates: 

Created #412 to address the need for all docs pages to be a part of the TOC somewhere.

Verified that `topicHref` is the way forward. Will do PRs against this issue that add `topicHref` to any article that links to another `toc.yml`. For starters I will have it link to whatever the first listed topic in the child `toc.yml` file is.

SeanKilleen:
This is completed via #413. Meant to associate it directly when merging. 👍

PR #413, opened 2020-06-25
Add topicHref items for child toc.yml entries

SeanKilleen:
Supports #350.


SeanKilleen:
FYI, I'm using regex search in vs code to find every `toc.yml` file that includes `toc.yml`, a newline, and then a new item (indicating it never had a `topicHref` underneath it). So I'm confident I'll find all the instance.

SeanKilleen:
Upon spot-check, this appears to accomplish what I'm looking to do -- searching for the main headings, e.g. `Running Tests`, now shows up in TOC search and allows a clickable link that goes to the first child topic.

Issue #419, opened 2020-06-26
build process: parallelize linting, spell check, and build


SeanKilleen:
While adding #418 I realized that the containers for spell check, build, linting, etc. are all being pulled sequentially. Adds time to the build process. I think we can do better than that.

PR #436, opened 2020-06-28
Parallelize the build process

SeanKilleen:
Splits the GitHub actions into four parts:

* Markdown linting
* Spellcheck
* Docfx build
* Publish (which depends on the prior three succeeding in parallel).

Process should now be much more responsive. Faster feedback for the win!


SeanKilleen:
Resolves #419.

Issue #441, opened 2020-06-29
Determine the best way to move forward on legacy docs publishing


SeanKilleen:
cc @jnm2 as FYI since he raised the original issue.
cc @rprouse as I believe it's his call ultimately.

## Background

In https://github.com/nunit/nunit.github.io/issues/79, @jnm2 realized a docs conflict caused the legacy docs to become unavailable.

## Problem 

This indeed appears to be a breaking change because GitHub's DNS and hosting is having trouble hosting both `nunit.org/docs/*` and `docs.nunit.org` due to the overlap of `docs` at that specific place in the hierarchy.

This is unfortunate; GitHub support confirmed that this should have been fine.

We resolved the symptom with #439 but now need to figure out the best way to proceed.

## Options

Rob is away at the moment but I can see a few courses of action here:

1. I could rename the docs repository to something like `thedocs` (to ensure it doesn't conflict with other similar terms), in which case GitHub would handle all the repo-level stuff for us and nothing should be an issue. I am not sure that it will solve the problem if GitHub's helpful handling of redirects keeps the issue in place.
2. I could do a quick PR capturing the old documentation and moving it to docfx as-is -- e.g. telling docfx to copy the raw html files and not modify them in any way. I believe the result will be that it looks as it does now.
3. Move the docs from the nunit site to a legacy, e.g. Charlie's site as he mentions below, and stop hosting them on nunit.org altogether (with redirect links, etc. maybe?) 
3. Something I haven't thought of?

## My suggestion

I think option 2, hosting the legacy docs in the new docs site, makes the most sense. However, in that scenario, I assume that the documentation is fully archived and does not need to be touched (I am using the Jekyll HTML output, not the jekyll inputs).

I will prepare a PR so that we are ready to go if Rob agrees with option # 2.

CharliePoole:
If it's of any use, at least on a temporary basis, the entire legacy docs are available on the NUnitSoftwarer site: http://www.nunitsoftware.com/nunitv2/index.php?p=documentation

@SeanKilleen Others know this, but for the "new guy"... the team decided to no longer maintain NUnit V2 a while back. I disagreed so i created the legacy project and site. That's open source for you! :smiley_cat:

rprouse:
Could we move the legacy docs to a different folder under the site? To be honest, I wouldn't mind breaking existing links to the legacy docs. I hate googling NUnit and getting links to the version 2.5 docs because they still have the most incoming links.

CharliePoole:
[...]

PR #442, opened 2020-06-29
Add Legacy NUnit Docs

SeanKilleen:
In support of #441. 

This adds all legacy documentation to the new docs site, so that it can live there if we choose as an option for resolving the conflict.

It should look and function as-is. I took the Jekyll output and copied it into the appropriate places along with the legacy files, css, images, etc.

On my local build of the docfx site, I now see what I would expect at the legacy path:

image


SeanKilleen:
Per #441, I'm merging these changes to add the legacy docs to our current site. This should enable us to publish the site again without breaking legacy links.

PR #447, opened 2020-07-07
Add missing NUnit 2.6.4 link to documentation index page

CharliePoole:
NUnit 2.6.4 was the last release of NUnit V2 made by the NUnit Project. On the original nunit.org documentation page it was shown as "Previous Release" but I don't think it needs to be highlighted that way any longer.

At the same time, I modified the text to clarify that we are talking about Legacy documentation.

@SeanKilleen I'm putting this through the PR process so you can use it as a way to decide what the review requirements for PRs should be for this repo.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -15,14 +15,15 @@ This web site contains the documentation for all active NUnit projects as well a
 * [Team practices](xref:teampractices) describe how NUnit works and how our teams work.
 * [Specifications](xref:Specifications) are descriptions of features we plan to add.
 
-## Older Releases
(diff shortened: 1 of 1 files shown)


SeanKilleen:
This makes good sense to me and is in line with what we talked about in our prior discussions. Thanks for submitting!

CharliePoole:
That's great. For the time being, shall we assume that you are approving all merges?

SeanKilleen:
@CharliePoole yes, thank you, I'd appreciate that as I get things set up for a smoother process and socialize the long term vision. Hopefully won't be long.

Issue #448, opened 2020-07-10
Full-text search does not currently display on live site


SeanKilleen:
Works locally: 

![image](https://user-images.githubusercontent.com/2148318/87156771-309c1700-c28b-11ea-90e2-18ee17ce8d91.png)

But not on the live site: 

![image](https://user-images.githubusercontent.com/2148318/87156790-3560cb00-c28b-11ea-8677-91fa29fde372.png)

SeanKilleen:
This issue is that the code to enable search looks for the contents of `<meta property="docfx:rel">`:

> ![image](https://user-images.githubusercontent.com/2148318/87159279-17956500-c28f-11ea-8d4f-09e5be868946.png)

However, this only exists locally for me:

> 
![image](https://user-images.githubusercontent.com/2148318/87159347-3136ac80-c28f-11ea-9cfd-104679e4184a.png)

And not on the published site:

> 
![image](https://user-images.githubusercontent.com/2148318/87159369-3dbb0500-c28f-11ea-85cc-a4a8b28c6992.png)

## Possible Culprits

* Publishing issue misconfiguration
* Template issue
* Build container discrepancy between my local environment

SeanKilleen:
I believe the issue is that our build system's docfx is several versions behind the version I'm using locally.

I've submitted https://github.com/nikeee/docfx-action/pull/7 to request the update; if that's not possible, I'll fork & bring in-house.

CharliePoole:
I assume the fallback is building your own container, not forking `docfx` itself.

SeanKilleen:
> I assume the fallback is building your own container, not forking docfx itself.

Correct -- I'm crazy, but not that crazy. 😆

SeanKilleen:
# Option 1

Fork the GitHub Action for building docfx into my `SeanKilleen` repo, iterate on it there, and use it in our build process.

SeanKilleen:
# Option 2

Fork the GitHub Action for building docfx into the `nunit-org`, iterate on it there, and use it in our build process.

SeanKilleen:
@CharliePoole @OsirisTerje @jnm2 @rprouse @ChrisMaddock @mikkelbu 

I'd like your opinion on the above (👍 / 👎)

I think we're going to need our own fork of the `docfx-action` project so that we can control docfx version numbers, have containers available when we need them, etc. 

Perhaps I can make it more extensible and then contribute that back upstream eventually and jettison the project but in the interim I'd like to have the tooling we need. 

[...]

PR #468, opened 2020-07-12
Switch to NUnit's docfx-action

SeanKilleen:
Uses our repo, which builds to our container.

Also uses the docfx version 2.561. as of the action version v1.1.0.

Supports #448.


SeanKilleen:
Nice! It worked.

PR #475, opened 2020-07-21
Fix link

ChrisMaddock:
I think this is the right way to fix this link?

I also removed the undocumented Internal Trace item from the menu. If we haven't got round to documenting it 5 years after release, it's probably not likely any time soon. 😅 

There's still a spec page, which we may move over at some point.

### DIFF
--- a/docs/articles/nunit/technical-notes/nunit-internals/NUnit-Internals.md
+++ b/docs/articles/nunit/technical-notes/nunit-internals/NUnit-Internals.md
@@ -3,7 +3,7 @@
 * [Architectural Overview](Architectural-Overview.md)
 * [NUnit APIs](NUnit-APIs.md)
   * [Test Engine API](Test-Engine-API.md)
-  * [Engine Driver API](Engine-Driver-API.md)
+  * [Engine Driver API](Engine-Driver-Api.md)
   * [Framework API](Framework-Api.md)
 * [Framework Design](Framework-Design.md)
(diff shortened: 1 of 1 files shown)


ChrisMaddock:
This is odd. I initially fixed the missing hyphen in the link, but CI failed. I then noticed the casing was also incorrect, so fixed that in the second commit, but CI's still failing.

CI on the second commit though seems to not have included the content of the second commit? (i.e. the link it is failing on is still showing in the incorrect case, rather than the correct casing I changed it to in that commit.)

And ideas, @SeanKilleen? Or is the casing actually a red herring, and there's something else wrong that I've missed?

image

ChrisMaddock:
Ah yes, I think something's up in CI. I think the screenshot below shows that, for my second commit (`d52b`), CI is checking out the code for the first commit (`2c881`). 

I think I'm reading that right? I hope so - otherwise I'm probably missing something much more obvious! 😅 

image

SeanKilleen:
@ChrisMaddock that's really odd...I want to say it's an issue with the GitHub action due to a merge commit but that is definitely strange, and it built fine on my local. 

I tried merging from master and pushing to the branch to see if that will fix things up. If that doesn't work, I'll rebase and force-push.

SeanKilleen:
Getting the update into the branch worked, I think it was a quirk. 🤷‍♂️ -- thanks for bearing with it, and for the update, @ChrisMaddock!

PR #476, opened 2020-07-22
Truly parallel builds (from an attempt at html-proofer)

SeanKilleen:
This started out as an attempt to use html-proofer, but then I realized it would need the artifacts.

So, this PR expanded a little.

It now:

* Splits the build & publish steps into separate steps
* Creates artifacts in the build step
* Uses those artifacts in the publish step
* Also added an html-proofer step that uses the build outputs.
* Runs everything in parallel as much as possible.

Supports #352

### DIFF
--- a/.github/workflows/build-process.yml
+++ b/.github/workflows/build-process.yml
@@ -61,7 +61,7 @@ jobs:
     needs: [build_and_publish]
     runs-on: ubuntu-latest
     steps:
-    - name: Download math result for job 2
+    - name: Download site artifacts
       uses: actions/download-artifact@v1
       with:
         name: _site
--- a/.github/workflows/build-process.yml
+++ b/.github/workflows/build-process.yml
@@ -47,6 +47,7 @@ jobs:
       uses: actions/upload-artifact@v2
       with:
         name: _site
+        path: docs/_site
   publish:
     name: Publish (master branch only)
     needs: [build]
--- a/.github/workflows/build-process.yml
+++ b/.github/workflows/build-process.yml
@@ -43,11 +43,13 @@ jobs:
       name: Build with Docfx
       with:
         args: docs/docfx.json --warningsAsErrors true
+    - name: zip site contents
+      run: zip -r _site.zip .docs/_site
     - name: Archive site artifacts
       uses: actions/upload-artifact@v2
       with:
-        name: _site
-        path: docs/_site
+        name: siteArtifact
+        path: _site.zip
   publish:
     name: Publish (master branch only)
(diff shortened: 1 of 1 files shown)


SeanKilleen:
Ah, right, I forgot -- I'm attempting to do this without any way to persist artifacts between builds. 

Ideally, what would happen is, in parallel:

* linting
* spell-checking
* building (and creating an artifact)

And then the publish step and the html proofer step download the artifact and use it if needed.

I'll expand this PR to that scope but it'll probably take a little longer to get it all sorted.

.github/workflows/build-process.yml
```diff
@@ -52,3 +57,15 @@ jobs:
         github_token: ${{ secrets.GITHUB_TOKEN }}
         publish_dir: ./docs/_site
         publish_branch: gh-pages
+  check_links:
+    needs: [build_and_publish]
+    runs-on: ubuntu-latest
+    steps:
+    - name: Download math result for job 2
```
SeanKilleen, line 64:
```suggestion
    - name: Download site artifacts
```

That's what I get for pulling from an example. 😆

.github/workflows/build-process.yml
```diff
@@ -44,11 +43,45 @@ jobs:
       name: Build with Docfx
       with:
         args: docs/docfx.json --warningsAsErrors true
+    - name: zip site contents
+      run: zip -r _site.zip .docs/_site
```
SeanKilleen, line 47:
```suggestion
      run: zip -r _site.zip . -i .docs/_site
```

.github/workflows/build-process.yml
```diff
@@ -52,3 +70,19 @@ jobs:
...
+    - name: Download site artifacts
+      uses: actions/download-artifact@v2
+      with:
+        name: siteArtifact
+    - name: unzip site contents
+      run: unzip _site.zip
+    - name: Check HTML
+      uses: chabad360/htmlproofer@master
+      with:
+        directory: "./docs/_site"
```
SeanKilleen, line 87:
```suggestion
        directory: "./docs/_site/articles"
```

SeanKilleen:
What started off as a request to use html-proofer will instead become "a bunch of stuff instead of html-proofer". 😆 

html-proofer is running, but it's leading to 429 errors because it's hitting GitHub links and getting rate limited, and I want to be a better internet citizen than that. 

I'll figure out a way to stem the issue, but for now will copy this config into #352 and remove it from here. 

Still merging the PR because we're getting awesome parallel builds at least.

PR #481, opened 2020-07-23
RegexConstraint doc update

### LINKED ISSUE #480: Very minor missing semi-colon in example
For the lazy, copy-pasters out there, like myself :smile:, on the RegexConstraint documentation page, there is a missing semicolon.

Also, another question - how many examples would you accept?

For instance, the RegexConstraint has the `IgnoreCase` modifier, but no example. Is having examples for such important, or being too pedantic?

I've pushed the two changes as separate commit so in case the answer is No, should be easy to revert.

Poimen:
Fixes #480


SeanKilleen:
Thanks @Poimen!

PR #496, opened 2020-08-02
Move legacy docs to "legacy" folder within source code

### LINKED ISSUE #492: Hide legacy docs in file hireachy
Making the recent change to the engine release notes, I got a little lost in looking for the latest version of the docs. My first thought when I opened the `/docs/` folder was to look for the version I want to edit for - I'm not sure it would be too clear to a new contributor that everything for version 3.x is under `articles` instead.

Is there a way we could push all those old docs down under `/legacy/` or something, and put in a redirect so they end up in the right place?

SeanKilleen:
Resolves #492.

## Work

* [x] Move legacy docs to `legacy` folder
* [x] Update docfx to grab them and publish them to the same destination as previously
* [x] Resolve the warnings that now appear on Index.md for a reference (warning on build; works when checking via `docfx serve`).

Incidentally, this also fixes a broken download link commonly found in legacy docs that I happened to notice.


SeanKilleen:
Update on this -- I am a little stuck with a docfx issue.

I am copying the docs, but am now seeing warnings in index.md around the references even though I know that's where the pages will exist after build.

It results in warnings only, but I don't want to loosen that build restriction.

SeanKilleen:
OK, I think I may have fixed it -- it's just the way I was referencing them needed to change. The URL in the end is exactly the same.

ChrisMaddock:
Sweet! Thanks for sorting this, Sean! 😄

PR #498, opened 2020-08-02
Add notes on --domain=None in Console Command Line

### LINKED ISSUE #43: Document how the --domain=None option works
Users need to understand the limitations of --domain=None so they will only use it if absolutely needed. See nunit/nunit#1409 for example. We should document this information somewhere.

SeanKilleen:
Resolves #43.


SeanKilleen:
Tagging @ChrisMaddock for review based on ownership and @CharliePoole based on context.

ChrisMaddock:
Looks good - thanks Sean! 👍

PR #499, opened 2020-08-02
Warnings around inheritance of attributes when hiding base method

### LINKED ISSUE #53: Base class SetUp method is not called if overridden in the derived class
I suspect this may be working as desired, but even so I'd suggest clarifying the documentation.

The documentation says

> You may define a SetUp method in the base class and another in the derived class. NUnit will call base class SetUp methods before those in the derived classes.

I’m looking at an existing codebase that uses that uses base classes to share common setup across many derived classes, like this:

```
public class Base
{
    [SetUp]
    public virtual void Setup() { ...

public class Derived : Base
{
    [SetUp]
    public override void Setup() { base.Setup(); ...
```

NUnit ignores the base class setup method, and only calls the derived class setup method.  (I know the derived setup calls the base setup, but I wanted to understand why _NUnit_ didn't call the base setup _first_.)

CommandBuilder.BuildSetUpTearDownList doesn’t just look for SetUp methods at each layer – it gets methods on the fixture (with Reflect.GetMethodsWithAttribute), then traverses the inheritance hierarchy to put them in order.  Since the derived class setup method _overrides_ the base class setup method, Reflect.GetMethodsWithAttribute only finds one setup method – the override on the derived class – it doesn't even see the base class setup method.

I’m guessing this is as-desired, so my suggestion is to simply add a small clarification to the documentation, something like this:

> You may define a SetUp method in the base class and another in the derived class. NUnit will call base class SetUp methods before those in the derived classes. If a base class SetUp method is overridden in the derived class, NUnit will not call the base class SetUp method.

Since it's been this way a while, I'd also suggest updating older versions of the documentation.

Note, if the derived class setup method _hides_ the base class setup method (either explicitly with ‘new’ or implicitly with no ‘new’/‘override’ modifier), then Reflect.GetMethodsWithAttribute finds _both_ setup methods, so it calls both, as expected.

[Tests to replicate.txt](https://github.com/nunit/docs/files/228732/Tests.to.replicate.txt)

Thanks,
Mitch

SeanKilleen:
Resolves #53.

Thanks to @mitchterlisner for reporting some years ago. 😄 We're getting our house in order so I thought I'd finally make these updates.

### DIFF
--- a/docs/articles/nunit/writing-tests/attributes/onetimesetup.md
+++ b/docs/articles/nunit/writing-tests/attributes/onetimesetup.md
@@ -51,6 +51,9 @@ You may define a OneTimeSetUp method
 in the base class and another in the derived class. NUnit will call base
 class OneTimeSetUp methods before those in the derived classes.
 
+> [!WARNING]
+> If a base class OneTimeSetUp method is overridden in the derived class, NUnit will not call the base class OneTimeSetUp method; NUnit does not anticipate usage that includes hiding the base method. Note that you may have a different name for each method; as long as both have the `[OneTimeSetUp]` attribute present, each will be called in the correct order.
+
 ## Notes
 
  1. Although it is possible to define multiple OneTimeSetUp methods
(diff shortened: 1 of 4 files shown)


SeanKilleen:
Tagging @CharliePoole and @rprouse to ensure I captured their thoughts correctly on this and it's stated in an accessible way. 

@mitchterlisner, as the original reporter, I welcome your feedback as well!

SeanKilleen:
@CharliePoole everything has a context. Best time to do so was back then; second best time is now. 😄

PR #500, opened 2020-08-02
Fix Specs menu item

ChrisMaddock:
This link is broken in the TOC. Is this the correct way to fix it, when a submenu is involved? I'm not totally sure, I'm afraid!

### DIFF
--- a/docs/articles/developer-info/toc.yml
+++ b/docs/articles/developer-info/toc.yml
@@ -1,5 +1,7 @@
 - name: Team Practices
   href: Team-Practices.md
+- name: Specifications
+  topicUid: Specifications
 - name: Notes Toward NUnit 4.0
   href: Notes-Toward-NUnit-4.0.md
 - name: Best Practices for XML Documentation
--- a/docs/articles/developer-info/toc.yml
+++ b/docs/articles/developer-info/toc.yml
@@ -1,7 +1,7 @@
 - name: Team Practices
   href: Team-Practices.md
 - name: Specifications
-  xref: Specifications
+  href: Specifications.md
 - name: Notes Toward NUnit 4.0
   href: Notes-Toward-NUnit-4.0.md
 - name: Best Practices for XML Documentation
--- a/docs/articles/developer-info/toc.yml
+++ b/docs/articles/developer-info/toc.yml
@@ -1,7 +1,5 @@
 - name: Team Practices
   href: Team-Practices.md
-- name: Specifications
-  href: Specifications.md
 - name: Notes Toward NUnit 4.0
   href: Notes-Toward-NUnit-4.0.md
 - name: Best Practices for XML Documentation


ChrisMaddock:
CI tells me that that was _not_ the right fix!

These pages are already linked under /NUnit/. Shall we just remove them from this menu?

SeanKilleen:
@ChrisMaddock let me check into this as broken links here should have generated a warning and not made it through the build.

What's the actual behavior you're seeing? The specifications link 404s?

ChrisMaddock:
This one seems to be generated as an empty link, with no href tag.

image

It's a second-level item under Developer Info. 🙂

SeanKilleen:
Interesting. The `Technical Notes --> NUnit Internals --> Specifications` link goes to https://docs.nunit.org/articles/nunit/technical-notes/nunit-internals/specs/Specifications.html, which has the `uid` of `Specifications`, meaning we should be able to reference it using `xref: Specifications` in the TOC. 

I'll play around with it and see what's up but I do see the issue you're describing.

SeanKilleen:
I figured it out; the issue is that it needs to be referenced via `topicUid` in TOC files; xref works for links in actual markdown pages. 🤷‍♂️

ChrisMaddock:
Great - thanks for digging in to it!

SeanKilleen:
Confirmed in my local that this is fixed. Merging. 👍

CharliePoole:
While "fixed" in the pure bug-fixing sense, the overall problem still remains. I wonder what could be done about it.

Although part of the documentation, these items are actually __not__ documentation. They are proposals for major additions to the framework and/or engine.

At the point when I left my role as the overall project lead, I was in the process of converting or deleting all the specifications that had driven development of NUnit 3. It was no longer my job but I didn't want it to be forgotten so I created that summary page with a table to track the removal of those items. As far as I can see, nothing further happened with them.

What ought to happen IMO is for each project lead to look at the items related to that project and make a decision. 

There are a few that are still interesting to me as part of my non-NUnit work, so I'll grab copies right now. :smile:

SeanKilleen:
@CharliePoole I treat a broken link and "what to do with items that may not actually be documentation" as separate items and suggest opening an issue to discuss it. 

I have some thoughts about these as well and I see it as part of a long term cleanup that I have yet to define, but your opinion on approach is always welcome.

CharliePoole:
@SeanKilleen This group has conducted many "post-mortem" discussions on issues after they are closed, leading to decisions to create new ones. That's what I was trying to accomplish here. Because it's a long-term habit of ours, I didn't bother stating the obvious (to me) point that the bigger issue is entirely separate and not something you could be expected to fix in this issue. :disappointed: 

If this goes as other such discussions have gone, we'll end up with one or more issues in the proper projects. I hope.

SeanKilleen:
@CharliePoole I prefer to have that discussion in a separate issue that references this issue to keep things clean. That discussion can then have agreed upon action items that come out of it. I'll probably add a bot to lock closed issues to make that approach less confusing / ambiguous. 👍

CharliePoole:
In that case, I'll stop doing it in this project!

SeanKilleen:
To be clear, I absolutely love the follow-up; I just prefer it to be in a new issue where it can be addressed holistically rather than an already-closed issue.

PR #508, opened 2020-08-09
Update description.md

CharliePoole:
The deleted text refers to a dialog that was part of the NUnitV2 GUI, which doesn't exist for NUnit 3.


SeanKilleen:
👍 Thanks!

PR #509, opened 2020-08-09
Update category.md

CharliePoole:
Removes references to NUnit V2 features: the GUI and the use of /include and /exclude command-line options. The framework should not try to specifically document any runners. It's enough to just say "some runners allow you to do this."


SeanKilleen:
Makes sense to me. 👍 Thanks @CharliePoole!

PR #510, opened 2020-08-09
Update test.md

CharliePoole:
Removed reference to the GUI which has not existed for a number of years.


SeanKilleen:
Thanks @CharliePoole!

PR #515, opened 2020-09-30
Error in --params description

bilbothebaggins:
The --params option was desribed with a pipe symbol instead of the required colon.

### DIFF
--- a/docs/articles/nunit/running-tests/Console-Command-Line.md
+++ b/docs/articles/nunit/running-tests/Console-Command-Line.md
@@ -36,7 +36,7 @@ Without the V2 driver, only version 3.0 and higher tests may be run.
 |`--test=FULLNAMES` | Comma-separated list of FULLNAMES of tests to run or explore. This option may be repeated. Note that this option is retained for backward compatibility. The --where option can now be used instead. |
 |`--testlist=FILE` | The name (or path) of a FILE containing a list of tests to run or explore, one per line. |
 |`--where=EXPRESSION` | An expression indicating which tests to run. It may specify test names, classes, methods, categories or properties comparing them to actual values with the operators ==, !=, =~ and !~. See [Test Selection Language](Test-Selection-Language.md) for a full description of the syntax. |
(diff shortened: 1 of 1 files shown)


mikkelbu (commented):
I think the intent is that you can either provide the `VALUE` using `--params` or using `-p` (so the pipe is an attempt to write a choice), but it seems that the documentation is not up-to-date wrt. the output from nunit3-console.exe which states that these options are deprecated
```
...
      --params, -p=VALUE     Deprecated and will be removed in a future
                               release. Please use --testparam instead.
      --testparam, --tp=VALUE
                             Followed by a key-value pair separated by an
                               equals sign. Test code can access the value by
                               name.
...
```
Wrt. `=` vs `:` then I don't think that this matter (but it has been a long time since I looked at this, so I might be wrong)

ChrisMaddock:
Hey @bilbothebaggins - thanks for flagging this, and going ahead with the change. I agree with Mikkel, the intention here is to show both the long and short options for this flag. We should be documenting both, but perhaps we can come up with a clearer way to show that?

The bigger issue here however is that `--params` is actually deprecated. We'd recommend using the `--testparam` flag instead - and this page really should be updated to reflect that.

Would you be willing to build on this contribution to reflect that? We'd really appreciate the help here.

CharliePoole (commented):
Sorry... just noticed the comment in the code I made some time ago is still pending. This should push it out to be seen.

General comments on format for command-line options...

* The __format__ in the docs should match the command-line help to avoid confusion.
* The __content__ in the docs should at least __start__ with the same text as the command-line help for the same reason, although it's possible the docs may need some additional info in some cases.
* The format at the command-line is driven by our use of Mono.Options, which actually generates the help text. Changing that would be a major rewrite of the options subsystem.
* @mikkelbu 's example shows the proper format, as generated by the Mono.Options code.

ChrisMaddock:
I think this has been superceeded by #555. Thanks anyway, @bilbothebaggins!

PR #519, opened 2020-10-20
Use nameof() in docs to reduce magic strings

svengeance:
I tend to view `nameof` as a step above strings in code for these kinds of things, however I understand this is _close_ to subjective preference. Feelings will not be hurt if this isn't the view of the maintainers, just thought I'd try.

This was a passing observation while reading docs, if there are other places you'd like this to be applied, let me know and I will.

Thank you~!

### DIFF
--- a/docs/articles/nunit/writing-tests/TestCaseData.md
+++ b/docs/articles/nunit/writing-tests/TestCaseData.md
@@ -10,7 +10,7 @@ The `TestCaseData` class provides extended test case information for a parameter
 [TestFixture]
 public class MyTests
 {
-    [TestCaseSource(typeof(MyDataClass), "TestCases")]
+    [TestCaseSource(typeof(MyDataClass), nameof(MyDataClass.TestCases)]
     public int DivideTest(int n, int d)
     {
         return n / d;


CharliePoole:
While this is a change I would advise in real code, I'm not sure it should be done in the docs.

My reservation is that it adds **One More Thing** to understand for the less-experienced programmer. The more-experienced programmer will quickly refactor to use of `nameof`.

It's just a reservation and I wouldn't argue strongly either way but I think it should be considered. IMO the framework team should think about their audience. In the past, it was an audience of experienced programmers and we deliberately did not try to teach them "best practices" in programming but only to show them how NUnit works.

As I say, I'm happy either way.

svengeance:
I understand where you're coming from. I guess as a counterpoint - when I'm learning how to program, and I'm looking up guides for the tools I'm using, **I would expect** that the guides follow the proper practices of the language. It would be embarrassing to copy-paste code from a highly respectable tool and have it fail a PR because there's a more standard way.  _But!_ I'm also far from an expert in this area of writing documentation, so it _may be the case_ that it is better to have triviality.

SeanKilleen:
First off, @svengeance, thanks for submitting the idea!

This is something that I've been torn on and am having difficulty making a strong argument for one way or the other. 

As I see it, arguments for:

* This really is the better way to do it in production code and is the way I'd recommend anyone to do it
* It also makes it clear to me that it's a reference to the test data below

Arguments for the string-based version:

* If someone isn't already using `nameof`, this may seem foreign
* As Charlie noted, people who are familiar with nameof can quickly refactor
* It's a little less verbose to refer to it by the string


I think where I ultimately come down on this is: it's a good idea. 

* C# 6 with support for `nameof` was released in 2015, half a decade ago. 
* The `nameof` concept is one that should be adopted widely and has little cost for someone to adopt. 
* The thing that put me over the edge is -- I think `nameof` is common enough that I'd rather someone ask the question and be able to point them to some tips, rather than do things in a less production-ready way to enable them to not have to ask that question.

Perhaps there's a middle ground where we can link to `nameof` documentation in the places we use it, but that may not even be necessary.

So while I think there are certainly arguments to be made to keep the string-based variation, I think I'm comfortable saying we should move forward allowing the use of `nameof`. 

I'll merge this, and perhaps file a follow-up in the docs. @svengeance, feel free to submit other PRs that continue this conversion, if that's something you're passionate about. 

Thanks for contributing this and kicking off the discussion!

svengeance:
@SeanKilleen Thank you! If you have an inclination to follow up on this with further work and file an issue, I wouldn't mind spending some time looking through the docs for consistency.

PR #524, opened 2020-11-13
Fixes #523 - finishes applying nameof() to areas where it can replace magic strings

### LINKED ISSUE #100: Add documentation for parallel tests
@CharliePoole commented on [Fri Sep 26 2014](https://github.com/nunit/nunit.org/issues/5)

### LINKED ISSUE #523: Finish adding nameof() to replace magic strings
This issue documents the continued effort started in #519 to substitute nameof() in places where magic strings are used -- usually in context of sourcing test case data.

Of note: while applying the fix for this, I discovered a couple places in the documentation where nameof is expressly recommended, namely [here](https://docs.nunit.org/articles/nunit-analyzers/NUnit1011.html#fix), [here](https://docs.nunit.org/articles/nunit-analyzers/NUnit1025.html#fix), and [here](https://docs.nunit.org/articles/nunit-analyzers/NUnit1021.html#fix).

svengeance:
Fixes #523 

This should finish the effort started in #519 .

Please advise if I've missed anything or if anybody feels there is an unjust replacement.

Thank you!

### DIFF
--- a/docs/articles/nunit/writing-tests/ListMapper.md
+++ b/docs/articles/nunit/writing-tests/ListMapper.md
@@ -13,9 +13,9 @@ shows two forms of the same assert.
 string[] strings = new string[] { "a", "ab", "abc" };
 int[] lengths = new int[] { 1, 2, 3 };
 
-Assert.That(List.Map(strings).Property("Length"),
+Assert.That(List.Map(strings).Property(nameof(string.Length)),
        Is.EqualTo(lengths));
 
-Assert.That(new ListMapper(strings).Property("Length"),
+Assert.That(new ListMapper(strings).Property(nameof(string.Length)),
        Is.EqualTo(lengths));
 ```
--- a/docs/articles/nunit/writing-tests/ListMapper.md
+++ b/docs/articles/nunit/writing-tests/ListMapper.md
@@ -13,9 +13,9 @@ shows two forms of the same assert.
 string[] strings = new string[] { "a", "ab", "abc" };
 int[] lengths = new int[] { 1, 2, 3 };
 
-Assert.That(List.Map(strings).Property(nameof(string.Length)),
+Assert.That(List.Map(strings).Property("Length"),
        Is.EqualTo(lengths));
 
-Assert.That(new ListMapper(strings).Property(nameof(string.Length)),
+Assert.That(new ListMapper(strings).Property("Length"),
        Is.EqualTo(lengths));
 ```
--- a/docs/articles/nunit/writing-tests/TestFixtureData.md
+++ b/docs/articles/nunit/writing-tests/TestFixtureData.md
@@ -10,7 +10,7 @@ The following example varies the example shown under [TestFixture Attribute](xre
 a `TestFixtureSourceAttribute` with a data source in a separately defined class.
 
 ```csharp
-[TestFixtureSource(typeof(MyFixtureData), "Fixtureparams")]
+[TestFixtureSource(typeof(MyFixtureData), nameof(Fixtureparams))]
 public class ParameterizedTestFixture
 {
     private string eq1;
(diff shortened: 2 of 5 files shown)


docs/articles/nunit/writing-tests/ListMapper.md
```diff
@@ -13,9 +13,9 @@ shows two forms of the same assert.
 string[] strings = new string[] { "a", "ab", "abc" };
 int[] lengths = new int[] { 1, 2, 3 };
 
-Assert.That(List.Map(strings).Property("Length"),
+Assert.That(List.Map(strings).Property(nameof(string.Length)),
```
SeanKilleen, line 16: I'd like to revert the instances in this file, as I think here it truly does make the understanding a little less clear at first glance.

docs/articles/nunit/writing-tests/TestFixtureData.md
```diff
@@ -10,7 +10,7 @@ The following example varies the example shown under [TestFixture Attribute](xre
 a `TestFixtureSourceAttribute` with a data source in a separately defined class.
 
 ```csharp
-[TestFixtureSource(typeof(MyFixtureData), "Fixtureparams")]
+[TestFixtureSource(typeof(MyFixtureData), nameof(Fixtureparams))]
```
SeanKilleen, line 13:
```suggestion
[TestFixtureSource(typeof(MyFixtureData), nameof(MyFixtureData.Fixtureparams))]
```

I believe this needs to include the class name, otherwise we'll see an error about it not existing.

docs/articles/nunit/writing-tests/attributes/testcasesource.md
```diff
@@ -17,10 +17,12 @@ Consider a test of the divide operation, taking three arguments: the numerator,
 
 ### Form 1 - [TestCaseSource(string sourceName)]
 
+#### Note: We use [nameof(DivideCases)](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/nameof) here to avoid introducing [magic strings](https://en.wikipedia.org/wiki/Magic_string) into our code, which offers better resilience when refactoring.
```
SeanKilleen, line 20:
```suggestion
> [!NOTE]
> We use [the `nameof` operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/nameof) to avoid introducing [magic strings](https://en.wikipedia.org/wiki/Magic_string) into code, which offers better resilience when refactoring. While `nameof` is recommended, you could also use the string "DivideCases" to achieve the same outcome.
```

Suggesting:

* Updated format (we have a concept for notes that works within docfx)
* Minor language changes
* Indicating what the string-based approach would be.

SeanKilleen (changes requested):
Looking great so far! I've left reviews suggesting some minor changes. Let me know when it's ready for another look.

docs/articles/nunit/writing-tests/attributes/testcasesource.md
ChrisMaddock: I like this suggestion! I can't make another suggestion on top of it, but could we also remove the `/en-us/` section of the docs.microsoft.com link? That means that users who normally use a translated version of docs.ms will be directed to the page in their default language, rather than specifically the en-us version. 🙂 

Thanks for doing this @svengeance - I think it's a great change overall. 😄

SeanKilleen: Great catch @ChrisMaddock -- I updated my suggestion so @svengeance can merge it in if they choose.

ChrisMaddock: Thanks Sean!

docs/articles/nunit/writing-tests/TestFixtureData.md
svengeance: Good catch, thank you!

docs/articles/nunit/writing-tests/ListMapper.md
svengeance: I can see both ways - for me personally, it's _nice_ that this reinforces the idea that "Length is a property on a string". I won't let it hold up anything - and I suppose my concern is partially remedied by the fact that the preceding method is `.Property(string property)`.

SeanKilleen (approved):
Looks great -- thanks again @svengeance!

svengeance:
Of course. Love NUnit. Thanks for the quick feedback.

PR #525, opened 2020-11-15
Standardize xrefs to lowercase

### LINKED ISSUE #513: Chore: Standardise format of xref's
Just done my first significant edit of the docs since we moved to the new system, which overall was a great experience! :smile:

One minor pain point I found was getting the correct case for my xrefs, as currently they are formatted inconsistently in the existing content. It would make editing the docs easier if there was standardised casing format throughout.

SeanKilleen:
Resolves #513.

* [x] Look up `xref:[A-Z]` in vscode search with case matching turned on. Searched in `.md` and `.yml` files
* [x] find/replace on the `xref`s
* [x] find/replace on the `uid`
* [x] find/replace on the `topicUid` in toc.yml files
* [x] Look up all uids and change them / topics

### DIFF
--- a/docs/articles/developer-info/toc.yml
+++ b/docs/articles/developer-info/toc.yml
@@ -1,7 +1,7 @@
 - name: Team Practices
   href: Team-Practices.md
 - name: Specifications
-  topicUid: Specifications
+  topicUid: specifications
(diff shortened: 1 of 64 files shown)


ChrisMaddock:
Great - thanks Sean! 😁

PR #532, opened 2020-12-02
feat(Analyzers): Add documentation for release 0.6

mikkelbu:


SeanKilleen:
@mikkelbu FYI, 2042 needs the code blocks to be updated to include the language, e.g. add `csharp` after the three backticks. That's what's failing the linter currently. Tried to make the suggested change but it didn't work on mobile at the moment.

PR #540, opened 2021-01-08
Make documentation up-2-date wrt. help information

mikkelbu:
Fixes nunit/nunit#3433

### DIFF
--- a/docs/articles/nunit/running-tests/NUnitLite-Options.md
+++ b/docs/articles/nunit/running-tests/NUnitLite-Options.md
@@ -7,7 +7,7 @@ The NUnitLite runner is invoked by executing the test program. If desired, any o
 | `--test=NAMES` | Comma-separated list of NAMES of tests to run or explore. This option may be repeated. |
 | `--testlist=PATH` | File PATH containing a list of tests to run, one per line. This option may be repeated. |
 | `--prefilter=NAMES` | Comma-separated list of NAMES of test classes or namespaces to be loaded. This option may be repeated. |
-| `--where=EXPRESSION` | Test selection EXPRESSION indicating what tests will be run. | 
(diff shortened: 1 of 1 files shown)


CharliePoole (approved):

CharliePoole:
LGTM One comment though...

The relationship among `--test`, `--where` and `--prefilter` isn't really clear from just reading this. What about a note following all the options explaining that?

SeanKilleen (approved):
@mikkelbu LGTM. Thanks for this!

Let me know if you want to address Charlie's recommendation as part of this PR or another; if you'd prefer to do it separately I'll merge this as-is.

CharliePoole:
@mikkelbu Would you like me to add something?

mikkelbu:
> @mikkelbu Would you like me to add something?

Please do. You are most welcome to add to this PR.

CharliePoole:
@mikkelbu Take a look at my additions and merge if it looks OK to you - or change as needed.

mikkelbu:
@CharliePoole I like the change. I was consider whether we should make it even more explicit that a consequence of prefiltering is that unspecified TestCaseSources, ValueSource ... are not run. I'm unsure if the regular user understands the exact consequences of "If specified, it limits the classes, which NUnit will examine to find tests.". By we can also merge this as it is, and then I'll create an issue for a page for prefiltering that we also can link to.

Ps. I cannot merge the PR as I don't have the rights (I think that only the administrators and @SeanKilleen can do this, at least I know that both I and @OsirisTerje don't have write rights). I'll ping @rprouse/you about this on slack.

SeanKilleen:
@mikkelbu yeah we'll need to get you those rights ASAP. Sorry it's taken so long.

For now, I'll merge. 👍

PR #553, opened 2021-03-27
NUnitLite doesn't support XSLT transforms

ChrisMaddock:
I'm fairly sure this is supported only in the Console. (The docs were originally copies of each other here.)

I can't find anything in the NUnitLite code which supports XSLT's


SeanKilleen (approved):

SeanKilleen:
Thanks @ChrisMaddock!

PR #556, opened 2021-04-05
Prepare docs for 4.0.0-beta.2

OsirisTerje:


OsirisTerje:
@SeanKilleen The spell checker is messing up.   It complains about "/home/runner/work/docs/docs/docs/articles/vs-test-adapter/Adapter-Release-Notes.md:14:53 - Unknown word (Pakrym)", .  That word is a name, and is in brackets, and starts with a capital letter.  "Also a performance improvement was done, thanks to [Pakrym](https://github.com/pakrym) for the [PR 821](https://github.com/nunit/nunit3-vs-adapter/pull/821)".  I've added it to the exclude list now, but we seriously need a better spellchecker.

SeanKilleen:
@OsirisTerje re: spellchecking, `Pakrym` is not a known word, so wouldn't it be expected that the spell-checker flag it until it is added to a list? 

Would seem we have a couple of options here:

* Continue as-is, adding words as they come up, which the spell-checker will alert us to and we can then ignore
* Exclude certain files from spell checker, such as release notes files
* Create some sort of regex that excludes usernames -- I believe cSpell would respect `@` followed by characters followed by a `[Space]`.
* Open to other options.

Thoughts?

PR #559, opened 2021-04-27
Changes for 3.13.2 release

rprouse:


cSpell.json
```diff
@@ -128,7 +128,9 @@
     "methodname",
     "netfx",
     "snupkg",
-    "testparam"
+    "testparam",
+    "gleb",
```
SeanKilleen, line 132: ℹ️ Since these are part of a username, I'd recommend moving them to the ignore list rather than the list of works, so that these spellings won't be auto-suggested if editing markdown in editors that support cSpell, e.g. VSCode. 

Really minor though and absolutely wouldn't hold up this PR over it.

SeanKilleen (approved):

cSpell.json
rprouse: @SeanKilleen isn't where I put them the ignore list? It is in the `ignoreWords` section not `words`?

SeanKilleen: Oh my goodness, you're totally right @rprouse -- sorry about that! I misread the location of your commit.

PR #582, opened 2021-08-30
Update Assert.True.md

Rares95:
Explain what the message parameter is used for.

### DIFF
--- a/docs/articles/nunit/writing-tests/assertions/classic-assertions/Assert.True.md
+++ b/docs/articles/nunit/writing-tests/assertions/classic-assertions/Assert.True.md
@@ -2,7 +2,7 @@
 
 **Assert.True** and **Assert.IsTrue** test that the specified condition is true.
 The two forms are provided for compatibility with past versions of NUnit and
-NUnitLite.
+NUnitLite. The message parameter is used for whenever the assertion fails.
 
 ```csharp
 Assert.True(bool condition);


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/nunit/docs?pullRequest=582) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/nunit/docs?pullRequest=582) before we can accept your contribution.<br/><br/>:x: Rares95 [sign now](https://cla.dotnetfoundation.org/nunit/docs?pullRequest=582)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/nunit/docs?pullRequest=582) it.</sub>

mikkelbu (commented):
I don't mind making it explicit what `message` is for, but then I think we should do it for all the 40 classical assertions (and perhaps also mention what `params` is for).

SeanKilleen:
Hi @Rares95! Thanks for stopping by and being willing to contribute to the documentation! Community contributions keep this running.

Quick question as I think about how to approach this - did you add these notes because you were looking for this information and it happened to be missing?

Asking because understanding the motivation will help here. I'm in agreement with @mikkelbu that if you think it would have been helpful to have, we can look into adding it to all of the assertions. If we were to take that approach, is that something you'd be open to expanding in this PR, if we were able to also provide guidance? 

Thanks!

Rares95:
Hi! Yes, basically would be great to have it for all the versions that use this.

I added this because I was looking for it and couldn't find what it does, and it isn't 100% clear. I also think we should explain what params do! I can definitely expand it to the others that might need it. However, I am unsure myself what do the params do, however. :)

mikkelbu:
I must admit I very rarely use `message` and `params`, but they are quite similar to how `string.format` behaves (if not the same).

Given a test like this
```csharp
        [Test]
        public void SomeTest()
        {
            int someValue = 14;
            int anotherValue = 28;
            bool someBool = false;
           Assert.True(someBool, "This can only happen when the value is {0} and not {1}", someValue, anotherValue);
        }
```
then it will fail with the following result
```
  Message: 
      This can only happen when the value is 14 and not 28
      Expected: True
      But was:  False
```

Rares95:
I see! Yeah that makes sense :)

I will look into improving the docs then in these areas, will need to map out all the variants that might use this. I will probably do it over the weekend :D

PR #588, opened 2021-09-09
Retroactively document breaking change in v3.5

### LINKED ISSUE #130: Update Breaking Changes with v3.5 distribution changes
Hi,

I was trying to use nunit 3.5.0 and noticed that nunit3-console.exe was not present in the downloaded zip package (https://github.com/nunit/nunit/releases/tag/3.5), there is just the nunitlite-runner.exe. For the console you now have to download another package.
The thing is, this is a breaking change (for our scripts) which is not mentioned is the docs.

Will the future versions add back the nunit3-console.exe runner or only the lite version will be available from now one?
Why the lite runner is included but nunit3-console was moved to another package?

SeanKilleen:
Resolves #130.

### DIFF
--- a/docs/articles/nunit/release-notes/breaking-changes.md
+++ b/docs/articles/nunit/release-notes/breaking-changes.md
@@ -28,6 +28,10 @@ uid: breakingchanges
 
 * The AssertionHelper class has now been deprecated. Users can use the [NUnit.StaticExpect](https://github.com/fluffynuts/NUnit.StaticExpect) library as a near drop-in replacement.
 
+## NUnit 3.5
+
+* `nunit3-console.exe` is no longer shipped with NUnit and must now be used as its own separate package, `nunit-console`. The console was separated from the framework to allow the two components to be developed separately.
+
 ## NUnit 3.4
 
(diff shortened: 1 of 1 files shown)


CharliePoole (commented):
Seems fine to me, although I haven't done such things retroactively in the past... and this is pretty old!

Note that this is in the NUnit framework docs even though it refers to the console. I'm not even a team member for that project although Chris is. I think you should ask Rob.

SeanKilleen:
Thanks @CharliePoole! Adding @rprouse.

rprouse (commented):
This isn't quite correct. The NUnit console was always in a separate package, the only thing that changed is that it stopped being released in the ZIP file that is downloaded from GitHub releases. We could document that but it was documented in the release notes here https://github.com/nunit/nunit/releases/tag/3.5 and here, https://docs.nunit.org/articles/nunit/release-notes/Pre-3.5-Release-Notes.html#nunit-35----october-3-2016. I'd say that is enough.

SeanKilleen:
Ok, sounds good. I'll close this PR and the associated issue.

PR #593, opened 2021-09-20
Update Console-Command-Line.md

CharliePoole:
Added notes to deal with a couple of common user misconceptions.

* Many users seem to assume that `--framework` will select a particular runtime in a multi-targeted project. Of course, it could have been designed that way but multi-targeted projects didn't exist at the time. :-)
* Another common mistake is to use `--x86` in an attempt to select the x86 platform. (I've also seen `--x64`, which isn't even an option.

Of course, the misconceptions arise from a desire to have the particular selection features. We should deal with that in 4.0.


SeanKilleen (approved):
LGTM. Thanks for adding the clarity!

PR #594, opened 2021-09-21
Use GitHub Action rather published docker image

SeanKilleen:
Our build process refers to a published docker image. 

Rather than that, we should be able to use the Dockerfile that is in the GitHub Action's repo itself. 

I think this will avoid needing to publish an image to Docker hub, and allow us to clean up that extraneous DockerHub account etc.

### DIFF
--- a/.github/workflows/build-process.yml
+++ b/.github/workflows/build-process.yml
@@ -39,7 +39,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         name: Check out the code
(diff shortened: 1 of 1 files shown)


SeanKilleen:
ℹ️ This increases the build time as now GitHub pulls the Dockerfile and builds it rather than pulling from Dockerhub. However, it only takes ~55 seconds and allows us to kill off a whole dependency on Dockerhub, So I'm OK with it and will merge this.

PR #606, opened 2021-12-12
Added release notes info for the engine to the adapter

OsirisTerje:


Arhell (commented):
cc @SeanKilleen

docs/articles/vs-test-adapter/AdapterV4-Release-Notes.md
```diff
@@ -11,7 +11,7 @@ This is a bug fix release, with the following fixes:
 
 ### Engine update
 
-The NUnit.Engine has been updated to version 3.13 in this release.
+The NUnit.Engine has been updated to version 3.13 in this release.  See [engine release notes](https://docs.nunit.org/articles/nunit/release-notes/console-and-engine.html) for details.
```
SeanKilleen, line 14: There are a few ways to reference things in docfx. In this case, I suggest seeing if we can refer to the document by its ID. That way, if the URL for the site ever changes, docfx will still generate the correct URL.

I don't have the syntax in front of me, but happy to look it up once I'm at a computer.

If the other URLs for release notes are like this currently, let me know and we can merge this as-is and address holistically.

SeanKilleen (changes requested):
One small piece I'd like to change or make an issue to capture for follow-up. Comment provided.

PS @Arhell no need to CC me. I receive an alert of every PR that comes to docs. 👍

docs/articles/vs-test-adapter/AdapterV4-Release-Notes.md
OsirisTerje: This is outside my knowledge :-)  So happy if you fix, but good if we merge this now, since this reference was missing, and there has been a lot of changes to the engine.

SeanKilleen: @OsirisTerje I'll fix it right now and merge it. 👍

SeanKilleen (approved):

PR #617, opened 2021-12-17
Adding information on new properties for runsettings in the adapter

OsirisTerje:


docs/articles/vs-test-adapter/Tips-And-Tricks.md
```diff
@@ -156,6 +157,8 @@ Using the runsettings should be like:
 </RunSettings>
 ```
 
+Note that the ```Where``` statement does not work for the Visual Studio Test Explorer, as it would generate a conflict with the test list the adapter receives. It is intended for use with command line tools, dotnet test or vstest.console.
```
SeanKilleen, line 160:
```suggestion
Note that the ```Where``` statement does not work for the Visual Studio Test Explorer, as it would generate a conflict with the test list the adapter receives. It is intended for use with command line tools, `dotnet test` or `vstest.console`.
```

Nitpick suggestion.

SeanKilleen (approved):
👍 LGTM. One nitpick suggestion to address as you see fit.

Issue #620, opened 2021-12-21
update current year in footer


Arhell:
![21-12-2021 20-37-32](https://user-images.githubusercontent.com/26163841/146981470-7cd67222-e2b0-4e82-90eb-5ea8685c9510.jpg)
cc @SeanKilleen 
js?
in the DocFX documentation, I unfortunately did not find how to do it differently

SeanKilleen:
@Arhell I'd say this is a lower priority and I'd be happy to update the hard-coded value. 

But yes, if we want to tackle this, JS would probably be the best way given Docfx's current constraints.

The current hard-coded value is found in `footer.tmpl.partial`.

OsirisTerje:
That year should **not** be updated.  It is the year of First Publication, and thus has to stay that way.   However, since the site is constantly evolving, one *can* add a range to it, so that it reads  2018 - 2021.  That is the format we use for the adapter.  If one choose not to use  a range, which one can, then the first date of publication should be used, thus 2018.

Arhell:
> That year should **not** be updated. It is the year of First Publication, and thus has to stay that way. However, since the site is constantly evolving, one _can_ add a range to it, so that it reads 2018 - 2021. That is the format we use for the adapter. If one choose not to use a range, which one can, then the first date of publication should be used, thus 2018. See e.g. [copyrightlaws.com](https://www.copyrightlaws.com/copyright-symbol-notice-year/)

Got it, thanks, I will do 2018-current

PR #621, opened 2021-12-22
update current year(copyright)

Arhell:
Resolves https://github.com/nunit/docs/issues/620


docs/custom_template/partials/footer.tmpl.partial
```diff
@@ -8,7 +8,12 @@
         <a href="#top">Back to top</a>
       </span>
       {{{_appFooter}}}
-      {{^_appFooter}}<span>Generated by <strong>DocFX</strong></span> | Copyright (c) 2018 The NUnit Project - Licensed under <a href="https://creativecommons.org/licenses/by-nc-sa/4.0/" target="_blank">CC BY-NC-SA 4.0</a>{{/_appFooter}}
+      {{^_appFooter}}<span>Generated by <strong>DocFX</strong></span> | Copyright (c) 2018-<span id="currentYear"></span> The NUnit Project - Licensed under <a href="https://creativecommons.org/licenses/by-nc-sa/4.0/" target="_blank">CC BY-NC-SA 4.0</a>{{/_appFooter}}
     </div>
   </div>
 </footer>
+
```
SeanKilleen, line 15: I think there might be a separate place for the footer script includes. Let me confirm that

Arhell: Updated.
Here is an example by which I corrected
https://github.com/dotnet/docfx/blob/dev/src/docfx.website.themes/default/styles/main.js

made a build, everything works.

SeanKilleen:
@Arhell Thank you for the contribution! Please update your PR description: 

> image

From `#620` to `Resolves #620` so that this PR will automatically close the issue when merged.

To update, you can use the menu to the side:

> image

I'll check out the code now.

SeanKilleen:
ℹ️ As a general side note / tip in case it's helpful: you may want to avoid force-pushing branches, as it can get you into some trouble. 

Many small commits is fine; in the end, we'll squash and merge it into one commit, so there's no need to force push.

SeanKilleen:
Ah, I apologize -- I just noticed this is a draft PR and not yet ready for review. I'll pause until you're ready. 👍

Arhell:
> Ah, I apologize -- I just noticed this is a draft PR and not yet ready for review. I'll pause until you're ready. 👍

Done

Arhell:
I usually add an issue link here. (If there is access and the ability to do it). Then, after the merge, the issue will automatically close too.

SeanKilleen:
> I usually add an issue link here. (If there is access and the ability to do it). Then, after the merge, the issue will automatically close too.

Ah, I see! Thanks for pointing that out to me. I'm not used to that being in the UI, and my muscle memory expects the text. It's fine as is. 👍

SeanKilleen (approved):
👍 Thanks for another contribution!

PR #624, opened 2022-01-05
Accessibility - add lang to html

Arhell:
resolve https://github.com/nunit/docs/issues/623

### DIFF
--- a/docs/styles/main.js
+++ b/docs/styles/main.js
@@ -1,5 +1,6 @@
 var currentYear= new Date().getFullYear();
 document.getElementById("currentYear").innerHTML = currentYear;
 
+// NUnit docs adds the lang attribute ourselves for accessibility reasons, as docfx does not do this out of the box.
 var htmlElement = document.querySelector("html");
 htmlElement.setAttribute("lang", "en");
--- a/docs/styles/main.js
+++ b/docs/styles/main.js
@@ -1,2 +1,5 @@
 var currentYear= new Date().getFullYear();
 document.getElementById("currentYear").innerHTML = currentYear;
+
+var htmlElement = document.querySelector("html");
+htmlElement.setAttribute("lang", "en");


SeanKilleen:
I'm a little leery of modifying this template because it serves as an override. If a future version of docfx is released and we update to it, the possibility exists that the layout will be incorrect because we're overriding their default.

I'm wondering if the better fix for this isn't an upstream one: modifying docfx to take the language from the docfx settings JSON file if one is specified.

Arhell:
> I'm a little leery of modifying this template because it serves as an override. If a future version of docfx is released and we update to it, the possibility exists that the layout will be incorrect because we're overriding their default.
> 
> I'm wondering if the better fix for this isn't an upstream one: modifying docfx to take the language from the docfx settings JSON file if one is specified.
> 
> With that said, I'll look at this a little more closely when I'm able to see if it's worth it.
> 
> How large would you say the impact to the accessibility of the site is by specifying this language attribute?

i did with js

SeanKilleen:
@Arhell curious -- does this actually add additional accessibility if it's appended via JS after the page is rendered?

Arhell:
> curious -- does this actually add additional accessibility if it's appended via JS after the page is rendered, vs being in the HTML at the time of render?

The check goes through the tool, which means ok

docs/styles/main.js
```diff
@@ -1,2 +1,5 @@
 var currentYear= new Date().getFullYear();
-document.getElementById("currentYear").innerHTML = currentYear;
\ No newline at end of file
+document.getElementById("currentYear").innerHTML = currentYear;
+
+var htmlElement = document.querySelector("html");
+htmlElement.setAttribute("lang", "en");
```
SeanKilleen, line 5:
```suggestion
// NUnit docs adds the lang attribute ourselves for accessibility reasons, as docfx does not do this out of the box.
var htmlElement = document.querySelector("html");
htmlElement.setAttribute("lang", "en");
```

SeanKilleen (approved):

SeanKilleen:
ℹ️ In the future, it would be helpful to not force-push branches. I was interested in having the commit where you used the master template around, because it would be great to use as an example when I open the issue in the docfx repository shortly. 

I squash merge all the commits here, so it's OK if there are many commits -- helpful to have a history of 1) updated master template 2) reverted that update 3) used JS.

Just a tip for the future, not an issue.

Arhell:
> ℹ️ In the future, it would be helpful to not force-push branches. I was interested in having the commit where you used the master template around, because it would be great to use as an example when I open the issue in the docfx repository shortly.
> 
> I squash merge all the commits here, so it's OK if there are many commits -- helpful to have a history of 1) updated master template 2) reverted that update 3) used JS.
> 
> Just a tip for the future, not an issue.

Of course, thank you.
Already done issue https://github.com/dotnet/docfx/issues/7799

PR #643, opened 2022-02-10
Update release-notes.md

CharliePoole:
Add versions 3.14.0, 3.15.0-beta1 and 3.15.0

### DIFF
--- a/docs/articles/nunit-engine/release-notes.md
+++ b/docs/articles/nunit-engine/release-notes.md
@@ -4,7 +4,7 @@ uid: consoleenginereleasenotes
 
 # Console and Engine Release Notes
 
-## NUnit Console & Engine 3.15 - Fegruary 10, 2021
+## NUnit Console & Engine 3.15 - February 10, 2021
 
 Final Release of NUnit ConsoleRunner 3.15.0. No changes from the beta1 Release.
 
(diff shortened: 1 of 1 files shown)


docs/articles/nunit-engine/release-notes.md
```diff
@@ -4,6 +4,43 @@ uid: consoleenginereleasenotes
 
 # Console and Engine Release Notes
 
+## NUnit Console & Engine 3.15 - Fegruary 10, 2021
```
SeanKilleen, line 7:
```suggestion
## NUnit Console & Engine 3.15 - February 10, 2021
```

Typo

SeanKilleen (approved):
Fixed some spacing and a typo; otherwise looks good to me! 👍 

I'll enable auto-merge.

CharliePoole:
Thanks Sean. I was unable to figure out what was being called out as an error.

BTW, I noticed one thing in doing this... my general practice is to avoid editing user-submitted issue titles for spelling etc. simply because they wrote the issue, not I. So when we do spell-correct on issue titles, they may end up different in the release notes and in GitHub itself. I can't think of any problem that will cause however.

SeanKilleen:
> I was unable to figure out what was being called out as an error.

I clicked on the failing GitHub action, which took me to the logs. Had to scroll up a bit because unfortunately the output is verbose. (I'll see if I can fix that.)

> my general practice is to avoid editing user-submitted issue titles for spelling etc.

In this case the typo was the name of a month in a heading. But you're right, it shouldn't cause an issue if editing in the docs.

PR #647, opened 2022-04-02
Update to docfx action v1.8.0 / docfx v2.59.1

SeanKilleen:


SeanKilleen:
It builds! But it looks like a misspelling got merged in via a PR pr so I'll fix that too.

PR #652, opened 2022-04-30
Update DictionaryContainsKeyConstraint.md

DejanPopovic1:
IDictionary type is <int, int> instead of <int, string>


SeanKilleen (approved):
Thank you!

PR #653, opened 2022-04-30
Update EmptyConstraint.md

DejanPopovic1:
Most examples in nunit/docs are self contained. Example amended to become self contained.


SeanKilleen (approved):
👍

PR #655, opened 2022-04-30
Update UniqueItemsConstraint.md

DejanPopovic1:
Edited example so that it is self contained as per prevailing examples in nunit/docs.


SeanKilleen (approved):
👍🎉

PR #656, opened 2022-06-12
Add information that Visual Studio Test Adapter supports VS2022

mjedrzejek:


SeanKilleen (approved):
Thanks for this catch, @mjedrzejek!

SeanKilleen:
Oops; looks like our build process needs to bump some versions of things to ensure our spell check can run correctly. I'll get a PR in for that and then will double back to merge this. 👍

SeanKilleen:
Alright, since #658 has been merged, I am re running the checks and suspect they'll complete fine. If they still fail, we'll bring the contents of the main branch in and that should fix it.

SeanKilleen:
OK @mjedrzejek, looks like we'll need to pull the latest code into your branch on order for it to work.

You'll want to fetch the latest from the upstream repository and merge the master branch into your patch branch. This way your branch will contain the updated build pieces as well (and thus pass).

I'm not sure what your level of familiarity is with such things, but I'm happy to guide you if that'd be helpful. Just let me know!

SeanKilleen:
Actually, I'm sorry -- that's silly. This is a tiny change and I've already reviewed it. I'll just merge; no need for you to go through all those hoops just because our build process broke.

Thanks for your contribution!

PR #660, opened 2022-07-01
Standardize linting tooling on asterisk as list style

SeanKilleen:
Previously, the linting rule defaulted to ensuring that list items were consistent within a doc.

This makes it explicit in our tooling that asterisks are the standard (as they're in 99% of the docs already).

CC @CharliePoole per our discussion

### DIFF
--- a/.github/linters/.markdown-lint.yml
+++ b/.github/linters/.markdown-lint.yml
@@ -22,6 +22,8 @@ MD024: false
 MD013: false
 MD026: false
 MD046: false # Mix of indented and fenced code blocks
+MD004:
+  style: asterisk
 # MD004: false                  # Unordered list style
 # MD007:
 #   indent: 2                   # Unordered list indentation
--- a/.markdownlintrc
+++ b/.markdownlintrc
@@ -2,5 +2,8 @@
     "MD024": false,
     "MD013": false,
     "MD026": false,
-    "MD046": false
+    "MD046": false,
+    "MD004": {
+        "style": "asterisk"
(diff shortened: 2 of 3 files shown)


SeanKilleen:
Looks like there was only one article not using asterisks. I updated the article to accommodate the rule.

SeanKilleen:
@CharliePoole, I think I'll merge this to represent the current state of things; the docs are defaulted to asterisks.

I'm 100% open to suggesting a change to a different default; feel free to open an issue to discuss. It may even be that tooling can do this for us in one shot. I am not very opinionated on what the standard should be, only that there should be a standard. (though I do get the sense that asterisks are the common default these days.)

CharliePoole:
As I stated privately, I do not favor forcing use of one or the other. It's your decision and it's your purview to make the decision, but since I haven't expressed a preference publicly, I'll do it here...

My rationale...

* MarkDown is very happy to accept asterisks and dashes for unordered lists... even mixed, like this list.
- In extensive coaching work, I have never seen programmers confused by this... as curly braces sometimes do confuse them.
* It's a bad practice to arbitrarily constrain programmers (or authors) when there is no need to do so as it reduces contributions.
* MarkDown is used on lots of projects and many of us switch context between projects several times a day.

BTW, I am in favor of asterisks over dashes in any markdown, which may be viewed by users in a simple text editor, CHANGE and README files, delivered with the software, fall into this category. However, in the case of our docs, the markdown is source code, which the users don't see.

PR #680, opened 2022-11-19
Adapter 4.3.1 release

OsirisTerje:


OsirisTerje:
@SeanKilleen  There are some strange spelling errors from the spell check action, can you figure out what those mean ?

SeanKilleen:
HI @OsirisTerje -- I'm not available right now but it seems like at first glance:

* `discoverability` is not recognized as a word, so you need to add it to the "words" section within cSpell
* Same with `Tracelog`
* `internaltrace` should probably be wrapped in backticks
* `Dalsbø` should be added to ignore words since it's a name.

I'll take care of these when I'm able to show you. I'll also update cSpell in this PR to be less verbose.

SeanKilleen:
@OsirisTerje actually, I was able to fix them now 👍

docs/articles/vs-test-adapter/Adapter-Installation.md
```diff
@@ -14,13 +14,13 @@ Go to your project root and where you want to add your project.
 
 Create a folder for the project with a name matching what your new test project should be named.
 
-Then run the command ````dotnet new nunit```
+Then run the command ```dotnet new nunit```.
```
SeanKilleen, line 17:
```suggestion
Then run the command `dotnet new nunit`.
```

ℹ️ For single line commands, one backtick is fine.

docs/articles/vs-test-adapter/Adapter-Installation.md
```diff
@@ -14,13 +14,13 @@ Go to your project root and where you want to add your project.
...
 
 You will now get a complete nunit test project with the same name as the folder.  You will also have a template unit test class there as a starter.
 
 If you have a Visual Studio solution file in the root folder, you can go there and add the new nunit csproj easily:
 
-Assume you are at a solution root, and you either have a solution file, or have just been adding it with ```dotnet new sln```
+Assume you are at a solution root, and you either have a solution file, or have just been adding it with ```dotnet new sln```.
```
SeanKilleen, line 23:
```suggestion
Assume you are at a solution root, and you either have a solution file, or have just been adding it with `dotnet new sln`.
```

docs/articles/vs-test-adapter/Adapter-License.md
```diff
@@ -2,7 +2,7 @@
 
 ## MIT License
 
-Copyright (c) 2011-2021 Charlie Poole, 2014-2021 Terje Sandstrom
+Copyright (c) 2011-2021 Charlie Poole, 2014-2022 Terje Sandstrom
```
SeanKilleen, line 5: 👍 nice catch. I wonder if I can automate that.

SeanKilleen (approved):
Just a few minor suggestions and then I'd say go for it! 👍

Issue #687, opened 2023-01-01
⚠️ Build Process & GitHub Codespaces currently broken


SeanKilleen:
It's been a long minute. Just noticed it's not working.

It's because the devcontainer references a docker container that we no longer publish. We moved to the GitHub Container Registry, but that also hasn't been published for a minute and so I think the container has since been removed due to things being idle. 

So, I will:

* Make a build in docfx-action which should allow something to be published (hopefully successfully)
* Update the devcontainer in this repo to use the GitHub container registry container.

SeanKilleen:
An update to this -- I now know a little more about what's happening, though I still am now clear on the exact steps to fix it.

* We haven't built the action containers for a while
* When we did, they were rebuilt 
* The latest versions of the base images contained a git release where a security patch was applied to fix a CVE
* Git now requires the `.git` folder to be owned by the same user
* (I think) the Dockerfile has a user that is different than the GitHub actions user -- hence the error about dubious ownership. 

We have a few routes we could go:

* We can attempt to mark the various directories as safe that we need for the CI and the Codespaces environment.
 * This is probably the best route. I'm just unsure as of yet which directories I need to mark, and at what levels (in the docfx & codespaces containers? In the github action itself?)
* We could change the ownership of the files themselves.
 * This seems hacky and like it might have other unintended consequences
* We could disable the security measure altogether. 
 * I'd rather not do that.

SeanKilleen:
OK, in the end, I:

* Updated the `docfx-action` container in that repo. 
  * I got lucky because docfx just released `v2.60.0-preview.2` which works as an xplat .NET global tool, so I could revise that container to be an Ubuntu-based .NET SDK container with `docfx` installed as a tool.
  * I also had to add a workaround in that container so that git assumes all folders are safe.
* Updated our dev container
  * Based it on our `docfx-action` new preview container, which is based on the `docfx` preview.
  * Drastically simplified adding node, yarn, npm

With that, I believe this is working, and will close this once I verify it myself.

SeanKilleen:
Need to make one more fix. The workspace is being hard-coded and not calculated so it will fail on GitHub Codespaces vs VS Code codespaces on a machine.

PR #688, opened 2023-01-01
Update devcontainer location

SeanKilleen:
Supports #687.

We were pulling from an older DockerHub image that no longer exists, rather than the GitHub Container Registry that our docfx-action builds use. This will fix that.


SeanKilleen:
Hmm...something is odd here. Going to try creating a new PR.

PR #689, opened 2023-01-01
Overhaul dev container to latest Ubuntu-based docfx-action container release

SeanKilleen:
* Pull from the correct github container registry location. 
* Update the devcontainer to be based on new docfx-action container, which uses Ubuntu "Jammy" based .NET SDK and docfx installed as a global tool.

Supports #687.

The good news is, both the docfx-action container and the devcontainer end up being greatly simplified by this work, and it seems to be fully functional for our purposes even using the docfx `v2.60.0-preview.2`.


SeanKilleen:
OK, this looks like the build is failing due to the hosted github runner now having an issue because of a git security patch that was released on Debian. 

This thread sheds some great light on it: https://github.com/actions/runner-images/issues/6775

I'm honestly not sure where to go from here just yet, and I wasn't anticipating having to go this deep on things, so I'm going to have to put a pin in it for now.

SeanKilleen:
I added the git config update to the upstream docfx-action container which I'm hoping will help.

SeanKilleen:
Oops, silly mistake: despite changing it in the upstream repo, I still need to release a version of the action and bump the number in our runner. 🤦‍♂️

PR #692, opened 2023-01-04
Switch to new container edition for builds

SeanKilleen:
Supports #687.

Docfx recently released 2.60.0-preview.2 which is now a cross-platform .NET Global Tool. This is very exciting and allowed us to re-do the docfx-action to use the .NET SDK Ubuntu-based container which I think is going to give us really nice flexibility in the future.

This PR is to adopt our new action, though it's still in preview, to work out any issues.

* [x] Change to docfx-action v2.0.0-preview.3 preview container

### DIFF
--- a/.github/workflows/build-process.yml
+++ b/.github/workflows/build-process.yml
@@ -39,6 +39,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         name: Check out the code
+      - name: Set ownership
+        run: |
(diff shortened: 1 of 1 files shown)


SeanKilleen:
Hmmm OK, even with an entirely new container, with very few modifications, I'm getting the same "dubious ownership" issue. 

Now that I have a way more minimal repro, I'll try to see if I can raise it elsewhere to get smarter people than me to help. 😅

PR #693, opened 2023-01-04
Update release-notes.md

CharliePoole:
Added release notes for 3.16.0 and 3.16.1

### DIFF
--- a/docs/articles/nunit-engine/release-notes.md
+++ b/docs/articles/nunit-engine/release-notes.md
@@ -124,7 +124,7 @@ and communicate with them over a TCP connection.
 A second major feature in this release is engine support for preemptive cancellation when the normal approach of requesting the test run to
 self-terminate doesn't work. This is an engine feature, available to any runners supporting cancellation.
 
-* [382](https://github.com/nu__Bugs__nit/nunit-console/issues/382) Only 1 amdgent running with multiple projects and --process=Multiple
+* [382](https://github.com/nu__Bugs__nit/nunit-console/issues/382) Only 1 agent running with multiple projects and --process=Multiple
 * [418](https://github.com/nunit/nunit-console/issues/418) <TestAssembly.dll>.config files not loaded when using an .nunit project file and --process=Multiple
 * [551](https://github.com/nunit/nunit-console/issues/551) Make engine easier to Debug
 * [609](https://github.com/nunit/nunit-console/issues/609) Run after Reload reports assemblies multiple times
--- a/docs/articles/nunit-engine/release-notes.md
+++ b/docs/articles/nunit-engine/release-notes.md
@@ -4,6 +4,69 @@ uid: consoleenginereleasenotes
 
 # Console and Engine Release Notes
 
+## NUnit Console & Engine 3.16.1 - January 4, 2023
+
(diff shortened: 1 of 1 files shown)


CharliePoole:
@SeanKilleen I created this using the "user mode" approach of merely editing the markdown. It has errors. I'm able to display and understand the errors, but it offers me no way to incrementally change what I've already put in.

I could, of course, start over completely and try to avoid the errors I already made while making no new errors. This doesn't seem very useful. As it stands, I see this edit approach as an obstacle which tends to make me avoid writing new documentation - something I always enjoyed doing in the past! In this case, however, it's the release notes for two new releases so even if it's unpleasant I'd like to get them in before leaving the project. Any suggestions as to how I might proceed?

SeanKilleen:
Hi @CharliePoole, I think we discussed this before but I can't find the notes so I'll replicate it here.

First, an FYI as an aside -- the docs build process is broken due to an upstream issue with git in GitHub actions. A security patch that appears to cause a breaking change. So no matter what, builds can't go out right now. I'm addressing this as a top priority and am looking into how other projects have done this.

The "markdown linting" and "spellcheck" action checks in GitHub are the ones you'll want to look at. If there's an error in anyone, you'll want to review it just like you would a failed test. It should be able to guide you to a resolution or give the appropriate hint to help you clear it.

In terms of editing, I do typically have a GitHub dev container available which would open the repository in a vs code window in the browser, complete with all of the lovely plugins we use to help us detect and fix those issues. That is broken due to the same build issue above. Were you to check out the docs repo and open it in VSCode, you'd see all of those plugins suggested for install. I can appreciate the frustration this might cause, but I see it as akin to automated tests for code; we want it to pass to ensure a consistent, quality output.

BUT, with that said, I don't ever mind doing review and finishing up those final tasks for any of the project members, especially when you're putting in so much of the content. Give me a :+1: on this if you'd like and I'd be happy to get this over the finish line and merged in.

SeanKilleen:
I see that the linting & spell-checking have now passed.

Given that there is still a build issue, I verified the build locally. I will merge this and then once the build issue is resolved, this will go out to production.

CharliePoole:
@SeanKilleen Thanks for your answer. I was able to see that there's a global problem beyond linting and spell checking, which you'll have to take care of. My own errors are fixed now so hopefully you'll merge this after it's done.

I had figured out how to find the branch you create (docs:CharliePoole-patch2) and I have been continuing to edit it online. That's not documented anywhere I can see and I can't see how a user would ever know what to do when faced with the email referring to all the errors found.

To be clear, let me define the terms I'm using. I see two ways of working in the docs. By "user mode" I mean simply clicking on the "Improve this" button and making changes. It seems to me that should not be displayed if the user is going to receive back a message, which requires developer-level knowledge to deal with.

I do have a bit of knowledge, so I was able to reason out as follows: he's doing a PR so there must be a source branch. Maybe I have access to edit that branch. Turns out that's the case but I don't think it's something everyone can figure out and I'm not everyone would have write authority over the branch.

What I think of as "developer mode" involves checking out the project, doing a feature branch, creating a PR etc. Your comment above refers to that approach, which I had initially used for the docs. I'm very familiar with how to do that but I stopped doing it because I wanted to exercise the "user mode" approach to see how it worked.

Issue #427, opened 2020-06-27
Future consideration: Snippets in docfx


SeanKilleen:
Not sure this is possible, but while linting hundreds of files, I wished that it was. :) 

Several areas in the docs are repeated -- for example, some of the text in the notes for various constraints.

For consistency etc, I wonder if it's possible to do includes within docfx so that content that changes can be updated. I believe this is likely possible. 

If it is, we can find the similar text and then extract the parts that are the exact same.

SeanKilleen:
I think this is possible using includes: https://dotnet.github.io/docfx/spec/docfx_flavored_markdown.html#block

SeanKilleen:
Tested -- I was able to do an include via `[!include[](~/snippets/snippet-name.md)]`. We'd likely need to exclude those snippets from markdown linting since they'd break rules we'd otherwise want in place.

SeanKilleen:
OK, at this point, most longer/complete code snippets have been modernized.

Still plenty left to go, but already paying dividends as I was able to modernize to recommended practices such as `Assert.That` where warranted, and fix some issues with tests themselves (compilation, clear up intent, etc.)

PR #702, opened 2023-01-06
Introduce programamtic snippets via docfx

SeanKilleen:
Supports #427. This initial commit allows us to begin moving toward programmatic snippets that can be verified as actual code.

* Introduces a `Snippets.sln` file
* Adds a unit test example with production code
* Adds regions around the sample
* Updates docfx to pull in those snippets -- in this case, for the `Multiple Asserts` article (chosen at random)
* Adds a build process for building & testing that Snippets solution.


SeanKilleen:
CC @ChrisMaddock, @OsirisTerje , @mikkelbu , @rprouse , @CharliePoole as I assume you might be interested in this for one reason or another.

.devcontainer/Dockerfile
```diff
@@ -19,5 +19,5 @@ RUN yarn -v
 # This is so that when someone updates their devcontainer name, it doesn't get screwed up
 RUN echo 'first_workspace=\"$(cd /workspaces && ls | head -1)\"; alias spellcheck="cspell --config /workspaces/$first_workspace/cSpell.json \"/workspaces/$first_workspace/docs/**/*.md\" --no-progress"' >> ~/.bashrc
 RUN echo 'first_workspace=\"$(cd /workspaces && ls | head -1)\"; alias lint="markdownlint -c /workspaces/$first_workspace/.github/linters/.markdown-lint.yml /workspaces/$first_workspace/docs/**/*.md"' >> ~/.bashrc
-RUN echo 'first_workspace=\"$(cd /workspaces && ls | head -1)\"; alias build="cd /workspaces/$first_workspace/docs && docfx build"' >> ~/.bashrc
-RUN echo 'first_workspace=\"$(cd /workspaces && ls | head -1)\"; alias serve="build && docfx serve _site -n \"*\""' >> ~/.bashrc
+RUN echo 'first_workspace=\"$(cd /workspaces && ls | head -1)\"; alias build="docfx /workspaces/$first_workspace/docs/docfx.json"' >> ~/.bashrc
```
SeanKilleen, line 22: ℹ️ Incidentally I realized our dev container syntax could be cleaner so I'm tacking it on here. Verified in my local devcontainer.

docs/snippets/Snippets.NUnit/MultipleAsserts.cs
```diff
@@ -0,0 +1,47 @@
...
+        Assert.Multiple(() =>
+        {
+            Assert.That(result.RealPart, Is.EqualTo(5.2));
+            Assert.That(result.ImaginaryPart, Is.EqualTo(3.9));
+        });
+
+        // Can also work with the classic assertion syntax
+        Assert.Multiple(() =>
+        {
+            Assert.AreEqual(5.2, result.RealPart, "Real Part");
```
SeanKilleen, line 42: ℹ️ This generates a warning which is showing up in our PR. I'll have to find a way to disable those when we're explicitly doing that. Won't be a problem, just a note to self.

PR #717, opened 2023-02-13
small header name change

OsirisTerje:


SeanKilleen:
@OsirisTerje our doc system allows us to actually create a warning box that will show up in yellow. Would that be preferred for this text? If so, I'll do a follow-up PR.

OsirisTerje:
Let's try it and see how it looks :-)

PR #723, opened 2023-02-13
Fix links discovered by using link-checking tooling

SeanKilleen:
Supports #352


SeanKilleen:
Realized that the link-checker library is a few years dead now it seems so I'm not going to automate this, but rather will merge this PR in without it so that we get the link fixes.

PR #729, opened 2023-02-22
chore: Correct class name

mikkelbu:
The class `TestCaseDataParameters` does not exist in NUnit. The https://docs.nunit.org/articles/nunit/writing-tests/TestCaseData.html page already use `TestCaseParameters`.

See also https://github.com/nunit/nunit.analyzers/issues/523#issuecomment-1439574581 for context


SeanKilleen (approved):
👍 Thank you!

PR #736, opened 2023-03-02
Create redirect for V3 adapter page

SeanKilleen:


SeanKilleen:
CC @OsirisTerje -- I've tried a few flavors of this but no luck. I think there may be a bug that's fixed in a later version of docfx, so I'm going to merge @Arhell's PR to update our docfx-action container to 2.62.1, get that into our main branch, and then see how this fares.

SeanKilleen:
With these commits, the redirect is now working. I also updated the V3/V4 page to include V4 in the title (it was only in the TOC).

PR #737, opened 2023-03-02
Update to docfx-action v2.3.0 (and thus docfx v2.62.1)

SeanKilleen:


SeanKilleen:
Did a visual check of the site and spot checks many pages. All looked good to me. Merging. 👍

PR #745, opened 2023-03-11
WIP: NUnit on-ramp minibook

SeanKilleen:
Going for it. Not time like the present. I imagine this will be in progress for a little while 😅 

Thoughts & feedback always welcome via comments as I go.

### DIFF
--- a/docs/articles/book/concepts/TestTradeoffs.md
+++ b/docs/articles/book/concepts/TestTradeoffs.md
@@ -1,0 +1,49 @@
+# Test Trade-offs
+
+Different types of tests have different trade-offs in their usage.
+
+Typically, automated tests are thought of as a pyramid or a funnel.
+
+* In a pyramid visualization, unit tests comprise the base of the pyramid (the largest part). On top of them are integration tests, then acceptance/functional tests, then UI tests. 
+* In a funnel visualization, the pyramid is inverted, and we think about unit tests as catching a majority of potential issues, followed by integration tests and acceptance/functional tets.
+
+The thinking behind both of these visualizations is that you want most of the tests in your project to be unit tests, followed by integration tests and acceptance/functional tests because of the trade-offs we're about to get into.
+
+While the approaches above offer a generalized, the reasoning behind them is the important part to consider.
+
+## The Journey Away from Fine-Grained Tests
+
+The further we move away from unit tests toward coarser-grained tests, a few things happen:
+
+* **Tests require more setup**. The further away you go from unit tests, typically the more setup those tests require. Instantiating a new database for each test run is much more setup than using a fake dependency.
+* **Tests can fail for more than one reason**. When a unit test fails, you almost always can pinpoint exactly what's happening. But an integration test may comprise many units; when it fails, how do you know which unit is responsible for the failure, unless you also have a failing unit test?
+* **Tests can fail for unrelated reasons**. For example, if you have a UI test that refers to a certain element, and the name or position of that element changes, your UI test may fail even though the UI is actually in perfectly working order.
+* **Tests take (much) longer to run**. As a rule of thumb, I can run approximately 500-1,000 unit tests per second. However, if I have to instantiate a real database and make round trips of data to that database, a given test will take substantially longer. I once worked on a project where a few thousand unit tests took a few seconds to execute, but a few hundred integration tests took a few hours.
+* **Tests take longer to fix**. Because of the longer execution time, multiple possible failure points, and potential for flakiness, troubleshooting these tests are often more difficult, which can lead to tem being left flaky for a long time or (worse) ignored or deleted altogether.
+
+With that said, having a project that completely neglects a layer is likely to suffer as well:
+
+* The team may make assumptions about how units of code work together, only to find that real components behave differently in practice.
+* The team may miss important considerations of coarser-grained tests, such as the contracts for an API
+* If there are no UI tests at all, it could be possible to deploy an application that passes and yet has a UI that is completely inoperable.
+
+## Finding the Right Trade-offs for Your Team
+
+Each codebase has a different context and set of trade-offs that might inform test types to use. Examples:
+
+* Teams with the ability to minimize the setup and execution time of their tests may benefit from more integration tests
+* Legacy projects with little test coverage often start with UI tests to establish a baseline of confidence, and then take some of those tests and "push them down" into several API tests or integration tests to alleviate some of the trade-offs
+* Teams with a high degree of non-coder collaboration may write a higher number of acceptance/functional tests because the language of the tests is closer to the language they use with their stakeholders.
+
+Keep some of the below in mind and you may avoid some pitfalls:
+
+* **Actively talk about and re-evaluate test types**. For example:
+  * If a number of UI tests have built up confidence and you've seen no failures, and those tests are appropriately covered by finer-grained tests, it may make sense to retire them. 
+  * If you keep getting caught off-guard by integration issues, it may make sense to invest more time in integration or acceptance tests.
+  * If you've discovered a way to reduce the execution time and maintenance burden of a given layer of tests, it may make sense to invest more in that layer.
+* **Remember: The goal is _confidence_**. 
+  * If a test fails, it should be treated as an issue until it can be proven otherwise. 
+  * Don't settle for flaky tests if you can at all avoid doing so. 
+  * If a test no longer serves to improve confidence in the system (and doesn't meaningfully play into the living documentation of the system), consider removing it or pushing it into finer grained tests.
+  * If the maintenance of a set of unit tests is costly and things are well-covered by integration tests that provide a high degree of confidence, perhaps some of those unit tests can be retired.
+* **Keep execution times as fast as possible**. The goal is to run as many tests as possible as often as possible. If a set of tests takes 6 hours to run, how will you be able to get confidence in pushing a branch of code prior to merging it in? More often than not, those tests will be skipped.
--- a/docs/articles/book/concepts/TestingConcepts.md
+++ b/docs/articles/book/concepts/TestingConcepts.md
@@ -1,0 +1,84 @@
+# Testing Concepts
+
+Lots of concepts are thrown around when we talk about automated testing. We'll briefly describe some of these here and then we'll work on applying them in the rest of the series.
+
+## SUT / CUT (Situation/Class Under Test)
+
+You might hear the term `SUT` or `CUT` or see these variables in tests you come across. Typically they mean the situation or class under test.
+
+## AAA: Arrange / Act / Assert
+
+Sometimes called the `AAA` approach, this refers to the way we might write a given automated test:
+
+* **Arrange**: Here, we set up the test for the action that will follow. This typically involves setting up the situation/class under test to take a particular action on it. Sometimes some or all of this arrangement is done by a `SetUp` method.
+* **Act**: In this step, we take an action. We call some part of our situation/class under test that either returns a value or that we expect to throw an exception.
+  * Typically, we try to keep the actions to a minimum -- preferably, 1. If you are testing different actions, that often means you're testing a different path through your code. In that case, we recommend creating two tests -- one for each path.
+* **Assert**: In this section of the test, we assert that a value is what we expect, or an exception has been thrown as we expect -- anything that indicates our expectations about the production code are met.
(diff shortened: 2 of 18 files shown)


SeanKilleen:
Leaving off here for a little bit. I left off on this challenge: I'd created some projects in the Snippets solution with the intent of building a small supermarket checkout system to demonstrate the evolution into state control, mocking, etc.

The challenge is: I want to show how these things evolve, but snippets can only pull from a file's current state. So I can't really do something like the [Santa Sleigh blog post I did](https://seankilleen.com/2021/12/santa-sleigh-nunit/), because it's not in its own repo where I can reference the commits and trying to do so would be messy.

So my choices are: write the code right in the markdown, or structure the snippets project so that there a lot of duplicate classes and structure so that I can pick the appropriate point-in-time snippet for the docs. This is further complicated because I can't just append `V1`, `V2` etc. because that would show up in any captured docs snippets that had the classname.

For now I'm leaning toward writing the code in markdown to get it to a good place (copying from VS), and then extracting into the snippets project once it feels more settled, to at least avoid some of the confusion.

docs/articles/book/concepts/TheWhy.md
```diff
@@ -0,0 +1,39 @@
...
+* **Design Drivers**. As we mentioned earlier, some benefits of tests are that they often keep code simpler and more modular, and that they force the developer to test-drive the usage of their code. TDD takes this to the next level -- from early on, you are engaging with your the shape of your code. As you approach the design of your code with TDD, it often becomes easier to spot dependencies as they emerge and make them explicit. By performing these cycles early and often, you gain many opportunities for insight into your code and its usage, and your design will benefit from that.
+
+## Why NUnit?
+
+NUnit is a testing framework for .NET. Like other frameworks -- xUnit, MSTest -- it provides a few core components that compose such a framework:
+
+* Ways to define tests
+* A runner to run the tests (though you can hook into the Visual Studio test runner, the ReSharper test runner, NCrunch, or any other runner of your choice)
+* An assertion library to use during testing (though you can substitute for [FluentAssertions](https://fluentassertions.com/) or the library of your choice)
```
OsirisTerje, line 32: I think it is important to point out that NUnit actually have two assertion libraries, the classic one, and the fluent one.  The "That" syntax and its siblings - the constraint model - is NUnits own fluent library.  
(Under the hood, the classic calls on the fluent, but that is for another section I would assume)

docs/articles/book/concepts/TheWhy.md
```diff
@@ -0,0 +1,39 @@
...
+
+We believe the concepts we'll discuss here will be able to be applied to any major test framework in most languages. If you couple the concepts in this guide with some documentation in any major test framework, we hope you'll be able to get where you need to be.
+
+So, the question becomes -- who will benefit from NUnit's particular _flavor_ of test framework?
+
+* NUnit benefits in particular from its longevity -- our first NuGet package was published in 2011, 13 years prior to this article being written. In that time, the library has amassed 200+ million downloads and has seen a ton of support from the community. To borrow the Farmers Insurance ad campaign, "We know a thing or two because we've seen a thing or two."
```
OsirisTerje, line 38: NUnit dates back to 1998 (I think), but picked up speed around 2002 with NUnit 2.2.  That means actually 25 years of experience :-)

docs/articles/book/concepts/TheWhy.md
```diff
@@ -0,0 +1,39 @@
...
+* We think that beginners and newcomers benefit from the _explicitness_ of NUnit and the way it approaches tests with attributes such as `[Test]`, `[SetUp]`, and `[TearDown]`. In some cases, xUnit relies on C# conventions in ways that might not be familiar.
```
OsirisTerje, line 39: Very much agree.  Also the "broadness", like the ability to add messages to Asserts (not in XUnit), the easy syntax for data-driven tests (TestCase, TestCaseSource), and extensibility - something that is not so much known, see a later section for more info :-)

docs/articles/book/getting-started/index.md
```diff
@@ -0,0 +1,198 @@
...
+* Adds a new test project using the `nunit` template, called `Conversions.Tests.csproj` (this will be our unit test project)
+* Adds a reference from the test project to the production code project so that we can see its contents.
+
+As for the `nunit` template, what does that do for us "out of the box"? In the `Conversions.Tess.csproj` file, we can see that a few libraries have been added for us:
+
+[!code-xml[PackageListing](~/snippets/book/getting-started/Converter.Tests/Converter.Tests.csproj#L11-L23)]
+
+* `Microsoft.NET.Test.Sdk` brings the test platform along and allows tests to be run using the `dotnet test` command. It's a good idea to add this to any .NET automated test project.
+* `NUnit` brings in the core NUnit library.
+* `NUnit3TestAdapter` surfaces NUnit3 tests for the Visual Studio / `dotnet test` runner. It essentially allows NUnit tests to be discovered.
```
OsirisTerje, line 54: > It essentially allows NUnit tests to be discovered.

It essentially allows NUnit tests to be executed.

(Discovery may or may not be happening with the adapter, so its main focus is to handle the execution)

docs/snippets/book/GettingStarted.Tests/UnitTest1.cs
```diff
@@ -0,0 +1,6 @@
+namespace GettingStarted.Tests
+{
```
OsirisTerje, line 2: Use file scoped namespace ?

docs/snippets/book/getting-started/Converter.Tests/TemperatureConverterTests.cs
```diff
@@ -0,0 +1,61 @@
...
+    public void ConvertCToF_37_Returns98point6()
+    {
+        // Arrange
+        var sut = new TemperatureConverter();
+
+        // Act
+        var result = sut.ConvertCToF(37);
+
+        // Assert
+        Assert.That(result, Is.EqualTo(98.6));
```
OsirisTerje, line 44: Since this is a decimal, the result may or may not match.  A good time to use the Within
```cs
Assert.That(result, Is.EqualTo(98.6).Within(0.1);
```

OsirisTerje (commented):
Great structure and setup.  Love this!

docs/articles/book/concepts/TestingConcepts.md
```diff
@@ -0,0 +1,88 @@
+# Testing Concepts
+
+Lots of concepts are thrown around when we talk about automated testing. We'll briefly describe some of these here and then we'll work on applying them in the rest of the series.
+
+## SUT / CUT (Situation/Class Under Test)
+
+You might hear the term `SUT` or `CUT` or see these variables in tests you come across. Typically they mean the situation or class under test.
```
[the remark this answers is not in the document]
SeanKilleen: I don't think that it needs to, and I've commonly taught it as situation because it helps students get in the mindset of thinking what they're actually testing, in which case system may be entirely too broad or confusing.

docs/articles/book/concepts/TestTradeoffs.md
```diff
@@ -0,0 +1,55 @@
+# Test Trade-offs
+
+Different types of tests have different trade-offs in their usage.
+
+Typically, automated tests are thought of as a pyramid or a funnel.
```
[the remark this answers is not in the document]
SeanKilleen: Definitely, that's in the plan 👍

docs/articles/book/concepts/TestingConcepts.md
```diff
@@ -0,0 +1,88 @@
...
+
+## DRY vs. DAMP
+
+Because tests are production code, the general programming concept of `DRY` (Don't Repeat Yourself) often comes into play with tests.
+
+However, tests have an important consideration -- they should be understandable quickly, with as little sleuthing as possible. If it serves the readability and understandability of the test, it's perfectly fine to repeat certain logic within a test rather than attempting to abstract it all away. In this way, we often say we prefer `DAMP` (Descriptive and Meaningful Phrasing) to `DRY`.
+
+Like many things, this is a balance that a team should attempt to find and actively maintain. Some parts of tests can be extracted into common methods; other abstractions might make tests too hard to understand. Be willing to re-evaluate these choices as projects evolve.
+
+## Test Coverage
```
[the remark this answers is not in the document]
SeanKilleen: I like this because it surfaces something that is a natural part of how I think but wouldn't be for others. I'll consider adding a section on "what do we test?" Noting that we want to test production code and want to avoid our tests mimicking production code with a lot of logic that itself can fail.

docs/articles/book/concepts/TestingConcepts.md
```diff
@@ -0,0 +1,88 @@
+# Testing Concepts
```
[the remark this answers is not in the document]
SeanKilleen: Absolutely. Love it. For me that's built into thinking about "smaller loops" but it deserves its own callout.

docs/articles/book/concepts/TheWhy.md
```diff
@@ -0,0 +1,39 @@
...
+* **Design Drivers**. As we mentioned earlier, some benefits of tests are that they often keep code simpler and more modular, and that they force the developer to test-drive the usage of their code. TDD takes this to the next level -- from early on, you are engaging with your the shape of your code. As you approach the design of your code with TDD, it often becomes easier to spot dependencies as they emerge and make them explicit. By performing these cycles early and often, you gain many opportunities for insight into your code and its usage, and your design will benefit from that.
+
+## Why NUnit?
+
+NUnit is a testing framework for .NET. Like other frameworks -- xUnit, MSTest -- it provides a few core components that compose such a framework:
+
+* Ways to define tests
+* A runner to run the tests (though you can hook into the Visual Studio test runner, the ReSharper test runner, NCrunch, or any other runner of your choice)
```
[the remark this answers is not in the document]
SeanKilleen: Agreed. This is trying to get at the fact that a basic attribute of a test framework is that it includes a runner of some kind, and that we do. Worth calling out that the primary way most use it is via the adapter and dotnet test though. 👍

docs/articles/book/concepts/TheWhy.md
SeanKilleen: I'm torn on this. My thinking:

* I want this guide to demonstrate a happy path
* I don't want to confuse people with too many concepts 
* The docs already exist for the classic assertions
* People should probably know they exist
* I'm under the impression that the core team (myself included) finds the constraint based syntax to be the recommended out of the box usage, since it also helps avoid explaining the need to not mix up expected and actual values.

So I'm leaning towards using the constraint syntax everywhere except maybe in a few cases when introducing asserts themselves.

docs/articles/book/getting-started/index.md
SeanKilleen: Good catch, thanks 👍

docs/snippets/book/GettingStarted.Tests/UnitTest1.cs
SeanKilleen: Torn on this. Not an easy way to demonstrate that that's what's happening in the examples so my concern is that people might copy/paste and be confused by the result.

docs/snippets/book/getting-started/Converter.Tests/TemperatureConverterTests.cs
SeanKilleen: Yes, technically. I think that gets into docs territory or another example scenario though. As is the test passes and AFAIK will always pass. Too many concepts to introduce for a "first use". 

Though I do think it speaks to the need for a "common pitfalls and remedies" article that maybe briefly delves into summarizing some of the analyzer checks and additional things to watch for.

docs/articles/book/concepts/TestingConcepts.md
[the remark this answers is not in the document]
OsirisTerje: Well, https://glossary.istqb.org/en_US/search?term=SUT   and also https://acronyms.thefreedictionary.com/SUT.  I don't think it is wrong to use situation, but I would add System to the list there.  It makes it inline with what I think is a more common use of the acronym

docs/articles/book/concepts/TheWhy.md
OsirisTerje: I absolutely agree with you about using the constraint model in all examples.  My point was just that the words "constraint model" is an NUnit technical term.  The words others would use, and which match more with how it looks and is used, is Fluent Assertions.  So when I teach testing, I say that "NUnit has its own fluent assertions" as a built-in and preferred syntax.   To me, you may just skip the classical all over :-)

docs/snippets/book/getting-started/Converter.Tests/TemperatureConverterTests.cs
OsirisTerje: Sure, when doubles are used in particular one should be aware that the EqualTo may not match, decimals I think are better, but not quite sure.  It all depends.  Perhaps have a pointer to a Tricks and Traps section for that?

OsirisTerje: And just for fun:  
image

PR #750, opened 2023-04-14
Update AnyOfConstraint.md

HeSitates:
The `AnyOfConstraint` does not work when the expected values is not a separate list of values of the same type as the value to test.


HeSitates:
@dotnet-policy-service agree

SeanKilleen:
Nice suggestion, @HeSitates -- thanks! Sorry I missed it back when you submitted it.

I'm going to clean it up a little bit and get it merged.

Also going to file a suggestion for the analyzer team, and will likely convert this into a snippet so that we check it regularly.

SeanKilleen (approved):

SeanKilleen:
Thanks again for the great contribution, @HeSitates! 🎉

PR #760, opened 2023-05-28
Update docfx-action to 2.5.0 in build/codespaces

SeanKilleen:
* [x] Updates docfx v2.65.3 -> v2.67.2
* [x] Add `_googleTagId` metadata with our existing tag
* [x] Reconcile all partial templates with existing templates (if only difference in header template is our google tag, remove it.)
* [x] Add "lang" metadata with property `en`
* [x] Addresses all breaking changes from docfx release notes https://github.com/dotnet/docfx/releases
* [x] Spot check in GitHub Codespace preview

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -15,7 +15,7 @@
 				"vsls-contrib.codetour",
 				"GitHub.vscode-pull-request-github",
 				"shuworks.vscode-table-formatter",
-				"ms-dotnettools.csharp",
+				"ms-dotnettools.csharp"
 			]
 		}
 	}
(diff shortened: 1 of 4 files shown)


.devcontainer/devcontainer.json
```diff
@@ -4,15 +4,20 @@
...
-		"oderwat.indent-rainbow",
-		"mdickin.markdown-shortcuts",
-		"davidanson.vscode-markdownlint",
-		"redhat.vscode-yaml",
-		"vsls-contrib.codetour",
-		"GitHub.vscode-pull-request-github",
-		"shuworks.vscode-table-formatter",
-		"ms-dotnettools.csharp"
-	]
+	"customizations": {
```
SeanKilleen, line 7: ℹ️ Noticed while working in dev container. New syntax now.

PR #777, opened 2023-07-22
Update docfx-action version

SeanKilleen:
v2.7.0 updates to the latest version of docfx


SeanKilleen:
✅ Checked this locally in a dev container. Everything works as expected -- snippets, includes, etc. as far as I can tell.

PR #782, opened 2023-08-03
Update docfx-action to 2.8.0 (docfx v2.70.0)

### LINKED ISSUE #781: Update docfx-action to v2.8.0 (docfx 2.70.0)
Notes here for my future reference:

* We'll want to double-check the theme locally and ensure our overrides that remain still make sense
* For convenience, we'll want to update the container's local `serve` command to add `--open-browser` per https://github.com/dotnet/docfx/pull/8934 
* Check the redirects and template usage of them, per https://github.com/dotnet/docfx/pull/8892

SeanKilleen:
Closes #781.

* [x] double-check the theme locally and ensure our overrides that remain still make sense
* [x] ~For convenience, we'll want to update the container's local serve command to add `--open-browser` per
https://github.com/dotnet/docfx/pull/8934~ This doesn't work in devcontainers
* [x] Check the redirects and template usage of them, per
https://github.com/dotnet/docfx/pull/8892


SeanKilleen:
✅ Checked redirects. They work.

❌ Tried to open the browser automatically but it doesn't work in devcontainers, which is fine; only offered minimal improvement over what we were doing.

✅ Theme and overrides look correct to me.

Going to go ahead and merge this.

PR #807, opened 2023-10-15
Update TestFixtureAttributeExamples.cs To give an example of using params in a constructor

Shiney:
see https://github.com/nunit/nunit/issues/1459 for when this was fixed


docs/snippets/Snippets.NUnit/Attributes/TestFixtureAttributeExamples.cs
```diff
@@ -37,6 +39,17 @@ public ParameterizedTestFixture(int eq1, int eq2, int neq)
...
 
+        // Can use params arguments (but not yet optional arguments)
+        public ParameterizedTestFixture(params char[] eqArguments)
+        {
+            _eq1 = eqArguments[0].ToString();
+            _eq2 = eqArguments[1].ToString();
+            if (eqArguments.Length > 2)
+                _neq = eqArguments[2].ToString();
+            else 
+                _new = null;
```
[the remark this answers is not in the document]
SeanKilleen: Fixed.

SeanKilleen:
It appears this is failing. @OsirisTerje do I need to bump the NUnit version to fix this or am I missing something?

OsirisTerje:
@Shiney Can you fix the tests please?   @SeanKilleen Don't believe this is the 4.0, but it should refer to 4.0.0-beta.1 for other examples.

Shiney:
@OsirisTerje How is this done? The problem seems to be that the two extra uses of TestFixtureData only work with a very recently merged change on the main NUnit repository but the snippets are built against 4.0.0-beta.1 ?

Shiney:
And there isn't a more recent beta nuget package

OsirisTerje:
Then add an Ignore attribute to it, with a text saying that is waits for the next version :-)

Shiney:
Hmm seems like it still fails to run the constructor regardless of any ignore attributes etc. on this, let's just wait until the next nuget package is used to progress this.

docs/snippets/Snippets.NUnit/Attributes/TestFixtureAttributeExamples.cs
```diff
@@ -37,6 +39,17 @@ public ParameterizedTestFixture(int eq1, int eq2, int neq)
...
 
+        // Can use params arguments (but not yet optional arguments)
+        public ParameterizedTestFixture(params char[] eqArguments)
+        {
+            _eq1 = eqArguments[0].ToString();
+            _eq2 = eqArguments[1].ToString();
+            if (eqArguments.Length > 2)
+                _neq = eqArguments[2].ToString();
+            else 
+                _neq = null;
```
manfred-brands, line 50: `_neq` declaration needs updating, append `?`, to allow assigning `null`

Shiney: Fixing

OsirisTerje:
@Shiney Did you  put the Ignore attribute on the class(Fixture) ?  If you do that, the ctor will NOT be called.

Shiney:
Yes. I still had the error. The code for checking if a constructor exists that we can call and actually calling it are in different places, so presumably I'm getting that error instead.

OsirisTerje:
@Shiney 4.0.0 and 4.0.1 is out now.  Can you check this PR again ?

Shiney:
Thank you

PR #818, opened 2023-10-20
Disabled the myget package feed so that dev packages dont get in the …

OsirisTerje:
…way of the beta.


SeanKilleen (approved):
👍 Makes sense. Thanks!

PR #827, opened 2023-11-04
Migration Guide for NUnit 4.0

### LINKED ISSUE #825: Create migration guide for upgrading to v4
There are two major breaking changes in v4.
We should create a clear migration guide to upgrade v3 code, with different usage, to a working v4 version of their code.

@nunit/framework-team VolunTears please :-)

manfred-brands:
Fixes #825


manfred-brands:
@SeanKilleen I used `~/images` for references to `.png` files as I saw this being done in other place, but these don't work in github mark-down.  There is one existing image reference in `assertions.md` that works in github.
It would be nice if it works there for all, but I don't know how that affects your deployment.

Arhell (approved):

SeanKilleen:
Beginning to review this now; will need to continue it tonight. @manfred-brands we can't guarantee that image references will work in GitHub markdown. The true  way to know if something will work is to serve the contents via docfx (for convenience, you could open the PR in a GitHub codespace and then type `serve` and that should fire it up. No worries though, I'll check it and change as needed.

SeanKilleen:
@manfred-brands @OsirisTerje how do we feel about possibly creating a "Migration Guide" in general and counting 3.x -> 4.x as a topic within it? I'm thinking this would then be appended or preprended to for future migration guidance that will inevitably occur. If you agree I'll make the changes.

SeanKilleen (commented):
(Moved comment to review pane for better discussion)

manfred-brands:
> we can't guarantee that image references will work in GitHub markdown
My question was about the relative path used. The images that use `../../images` would work, but the `~/images` not as in the repository the `images` directory is not in the root, but under `docs`. If the relative links work in `docfx`, I would prefer to use those so users can look in github.

manfred-brands:
> how do we feel about possibly creating a "Migration Guide" in general 

We should only need one for major (and incompatible releases).
I don't mind either way, as long as the most resent one is on top so user don't have to scroll over no longer relevant items.

docs/articles/nunit/release-notes/Nunit4.0-MigrationGuide.md
```diff
@@ -0,0 +1,169 @@
...
+global using StringAssert = NUnit.Framework.Legacy.StringAssert;
+global using DirectoryAssert = NUnit.Framework.Legacy.DirectoryAssert;
+global using FileAssert = NUnit.Framework.Legacy.FileAssert;
+```
+
+Note that this doesn't mean you have to target .NET 6.0. This also works if targeting .NET Framework as it is purely done on the source code level.
+
+## Assert.That with _format_ specification and `params` overload conversion
+
+These overloads were removed to allow for better messages in case of failure. See [The "Towards NUnit 4" article](../Towards-NUnit4.md#improved-assert-result-messages) for more information.
```
manfred-brands, line 141: You've made this one very long line.
Maybe a lint setting for a maximum line length would be good.

docs/articles/nunit/release-notes/Nunit4.0-MigrationGuide.md
```diff
@@ -0,0 +1,169 @@
...
+    private static int SomeCalculation() => 42;
+}
+```
+
+### Convert Classic Assert to Constraint Model
+
+Although the code can be converted manually, that is a lot of work.
+
+Luckily, the [NUnit.Analyzer](https://www.nuget.org/packages/NUnit.Analyzers) has had rules and associated code fixes for a while now.
+Version _3.10.0_ knows about the 2nd non-backward compatible change and will convert the _format_ specification and `params` into a `FormattableString`.
```
manfred-brands, line 43: @mikkelbu You might want to release that version before this is merged/published.

SeanKilleen:
@manfred-brands 

> If the relative links work in docfx, I would prefer to use those so users can look in github.

Ah, I see! Makes sense. I'll fiddle with that a bit.

> Maybe a lint setting for a maximum line length would be good.

In the docs, I've taken to keeping things on one line until they are separate thoughts/components, since they render the same. I could be a little more consistent, but that's the idea. Does that cause any difficulty for you that I might be missing?

docs/articles/nunit/release-notes/Nunit4.0-MigrationGuide.md
SeanKilleen: Good catch; I can hold off on this until I hear from @mikkelbu.

docs/articles/nunit/release-notes/Nunit4.0-MigrationGuide.md
```diff
@@ -0,0 +1,169 @@
+---
+uid: nunit40migrationguide
+---
+
+# NUnit 4.0 Migration Guide
```
SeanKilleen, line 5: @manfred-brands @OsirisTerje how do we feel about possibly creating a "Migration Guide" in general and counting 3.x -> 4.x as a topic within it? I'm thinking this would then be appended or preprended to for future migration guidance that will inevitably occur. If you agree I'll make the changes.

manfred-brands: As said, I don't mind.

SeanKilleen:
@manfred-brands image paths updated and appear to be shown correctly in both my docfx preview and GitHub markdown. Thanks for that; I'll apply it elsewhere.

manfred-brands:
> In the docs, I've taken to keeping things on one line until they are separate thoughts/components, since they render the same. I could be a little more consistent, but that's the idea. Does that cause any difficulty for you that I might be missing?

Not all software wraps long lines:

Github  Extensions:
image

Visual Studio:
image

Even though github (online) wraps lines, diffs are on a per line basis. If such a line is extremely long it look like a big change.

I prefer a reasonable maximum line length somehow.

SeanKilleen:
@manfred-brands  

> I prefer a reasonable maximum line length somehow.

Created #830 to discuss. Input welcome there! It never bothered me; I just scrolled to the right as needed. But I can definitely see the value of doing it.

docs/articles/nunit/release-notes/Nunit4.0-MigrationGuide.md
SeanKilleen: I decided the other contents were valuable enough and that this could be a living guide, so I didn't want to rush @mikkelbu before merging. If we get any questions on it, it can be answered easily eonugh.

mikkelbu: I'll see if I can get an release out tomorrow (work is taking most of my time/energy at the moment)

PR #828, opened 2023-11-07
Correct Pairwise output order

CoenraadS:
Now the documentation is in the same order the output is:

image


CoenraadS:
@dotnet-policy-service agree

SeanKilleen (approved):
Works for me 👍 Thanks for the contribution, @CoenraadS!

Issue #834, opened 2023-11-09
Document added support for IAsyncEnumerable on *Source attributes


stevenaw:
Support for was added in https://github.com/nunit/nunit/pull/4538 and https://github.com/nunit/nunit/issues/3391
For the following return types:
- `IAsyncEnumerable<T>
- `Task<IAsyncEnumrable<T>>`

On the following attributes
- `TestCaseSource`
- `ValueSource`
- TestFixtureSource`

Issue #756, opened 2023-05-21
Document that *Source attributes permit Task<IEnumerable> return types


stevenaw:
Method-based *Source attributes gained support for async return types in upcoming v4 and v3.14 framework versions via https://github.com/nunit/nunit/issues/2729
It would be good to document this change. List of updated attributes can be found in https://github.com/nunit/nunit/pull/4389

stevenaw:
@SeanKilleen Curious, how would something like this be approached for unreleased changes? Is there a "prerelease" section of the docs, or should a PR be put up and just not merged until the framework changes are released?

OsirisTerje:
@stevenaw FWIW: I have been in the same situation for the adapter, and I chose to add a note to indicate from which version the feature was available.   E.g. see  https://docs.nunit.org/articles/vs-test-adapter/Tips-And-Tricks.html#where

For future features I know which version it goes into, so I add that upcoming version number.

SeanKilleen:
@stevenaw  yep, @OsirisTerje is correct here -- I'll see if anything changed but docfx is a little clunky here; it doesn't easily allow multiple versions of docs AFAIK.

stevenaw:
Thanks @OsirisTerje @SeanKilleen !
I'll give that a try.

stevenaw:
Spare time has been a bit hard to come by lately. I'm going to remove my assignment from this for now in case anyone else wishes to pick this up

PR #836, opened 2023-11-11
A few points for Task<> and IAsyncEnumerable<> support on *Source attributes

stevenaw:
Relates to https://github.com/nunit/docs/issues/834 and https://github.com/nunit/docs/issues/756

A few quick notes about supported signatures for async method-sourced *Source attributes to start getting these into the docs. I'm only describing this as "Related to" instead of "Fixes" as I think some example method signatures could be helpful to add later.


Arhell (approved):

SeanKilleen (approved):
Great note, thanks!

PR #840, opened 2023-11-26
fix: Correct typo (missing semicolon)

mikkelbu:


manfred-brands (approved):
It would be great if such items could be detected in builds, but snapshot examples are often partial pieces of code.

OsirisTerje:
It can, if the code is placed in the snippets parts.  But, I see it is more suitable for longer examples.

SeanKilleen:
These will eventually be moved to snippets. For now, thanks for this fix!

PR #843, opened 2023-11-27
Generate API docs from NUnit source code

SeanKilleen:
NUnit 4.0 has me feeling bold. :) Figured I'd take a first shot at pulling in API docs from NUnit's xmldoc comments, as docfx can do.

The docs, as we know, are not in the same repo as the code, so our build process and any manual experimentation will have to copy some files into a folder in the docs project.

## Work

* [x] Run the process manually
* [x] Ignore directories so they don't wind up accidentally committed
* [x] README reference for local steps
* [x] Build process to generate API
* [x] Make it look nice with the existing contents
* [ ] ~~Link to some kind of API landing page (reference some other repos that use docfx to see how they do it)~~ 
  * Unnecessary -- API reference jumps right into the APIs, and that's fine for now.
* [x] Some kind of preview functionality...like NETLIFY PREVIEWS! 🎉 
* [x] Some kind of nice tooling for the dev container to pull stuff in easily when working on it locally.
* [ ] ~~Have it always pull from the latest NUnit release, semver-wise.~~
  * I'm going to punt on this one -- doesn't need to happen right away.
* [x] ~~Filter out the internal namespaces (get input from Core team)~~
  * Addressing as follow-ups
* [x] Surface a preview link in a PR comment
* [x] Necessary metadoc (readme, contributing) updates


SeanKilleen:
CC @OsirisTerje  -- FYI, I'm getting started on this. :)

SeanKilleen:
OK, at this point I can get it building locally and via GitHub actions with the v4.0.0 release. 🎉 

Next up: Make it look nice and incorporate it as I've seen other sites do. Going to take a look around and come back to this.

SeanKilleen:
I'm going to sleep on this and return to it.

Some issue with not being able to find the index.md file that seems to be right there when I look at it. 

Also, might need to check out some of the exclusion rules to ensure that file isn't being overwritten because of the metadata generation.

SeanKilleen:
Very cool - at this point we have build previews via Netlify! I have to work on the DX of surfacing them, but it works. 

The deploy preview as of now can be seen at https://app.netlify.com/sites/resilient-maamoul-8b1e0d/deploys/65697184f5a04e595f6d9ddf 

@OsirisTerje , check out the API reference being generated from the xmldocs of the published NUnit release!

image

SeanKilleen:
@nunit/core-team I wanted to get your feedback on this before pushing it through too much further. Please see the preview link and check out "API Reference".

I think it makes sense to publish this sooner than later and then pay some attention to beefing up the xmldoc comments in the future as we go. I like the idea of putting the inner workings, classes, members, method calls etc. in the API reference docs and moving some of the more technical docs references to point to that, thus freeing up more space for conceptual docs and articles.

OsirisTerje:
It looks good to me, but I don't really understand the process.  It seems the APi reference contains more text than the XmlDocs, and that is fair, but exactly how does this work?

mikkelbu:
@SeanKilleen I agree with your approach about "publishing sooner than later" and also the part about moving low-level technical stuff to the API.

@OsirisTerje On the pages I looked the text seemed to fit (of course there are also reference to base classes in .NET but that is to be expected). Can you give an example on the "more text"?

manfred-brands:
@SeanKilleen I like the link with the source code. It might mean that we need to be a bit more conscientious with the actual XML docs.

A few remarks:
I find it annoying that as soon as I click on a link in the TOC, the whole page refreshes and the TOC on the left hand side collapses. The current doc site keeps it open. Is that a parameter you can set?

There area lot of classes/interfaces that are for nunit contributors than for the nunit users, e.g. [IMethodInfo](https://65697184f5a04e595f6d9ddf--resilient-maamoul-8b1e0d.netlify.app/api/nunit.framework.interfaces.imethodinfo)
They might now be `public`, but maybe they should have been `internal`.

We do have entries for the `Internal` namespace, do we want to expose those?
If so what is the point of `internal`? I know some of the classes are public yet 90% of the users won't need those.
If you want to keep them could we move them to a separate section for constraint implementers?
Don't get me wrong, documenting some of the internal workings and classes would help future contributors.
But I do think it is separate from the nunit api?

OsirisTerje:
I clicked the links from @SeanKilleen , and that took me to the standard docs, no API under there, afaic see.  

The link from @mikkelbu  however, is completely different, and makes more sense, but I still can't find it in the other links.

Having these API references are awesome, but for e.g. the constraints, we need the standard documentation too, otherwise all documentation moves over to the code repo, and I assume that is not the intention.

Also, it seems that some information is leaking over, e.g. https://65697184f5a04e595f6d9ddf--resilient-maamoul-8b1e0d.netlify.app/api/nunit.framework.legacy.classicassert Classic Asserts, the Assert.That is also present there.  I assume that is something about the filtering.

But all in all, this starts to look good.  We need to be more careful with the in-code headers of course, but that is just fine.

SeanKilleen:
@OsirisTerje @manfred-brands @mikkelbu awesome feedback, all -- thank you! I'll try to address it collectively here.


@OsirisTerje 

>  It seems the APi reference contains more text than the XmlDocs, and that is fair, but exactly how does this work?
> Assuming I am looking at the right place

You were not looking at the right place. 😆 At the top of the preview site is a new "API Reference" link -- that's what you'll want to click.

image

> exactly how does this work?

docfx as a tool is designed to support both articles (what we currently do) and doc generation from the xmldoc comments. This is how docfx surfaces that. What I'm doing behind the scenes for a build is pulling the latest NUnit release and extracting it to a place where docfx can look at the xmldoc comments and thus generate this section of the site.

@manfred-brands 

>  It might mean that we need to be a bit more conscientious with the actual XML docs.

Agree. That's one of the reasons I wanted to be clear. We'd probably start getting more PRs for the xmldocs in NUnit too though, which could be a good thing. I'll also need to brush up on my xmldoc skills in terms of referencing other things etc. but there are some great repos out there we can reference.

> I find it annoying that as soon as I click on a link in the TOC, the whole page refreshes and the TOC on the left hand side collapses. 

Oh yeah, that's weird. Not sure if it's just a function of the Netlify preview or not. I suggest adding a follow-up task to check and remediate this as necessary.

> They might now be public, but maybe they should have been internal.

Great point -- I have a lot of flexibility in terms of which stuff to exclude, and can exclude anything we'd like. I'd love this team's input on what I should start with there -- perhaps excluding all internal classes and also certain additional namespaces / classes that don't make sense to expose publicly (would cause confusion, etc.). We can draw whatever line we want and change it as we see fit. With that in mind, it probably makes sense to exclude more things at the outset and then add them later if desired.

@OsirisTerje 

> constraints, we need the standard documentation too, otherwise all documentation moves over to the code repo, and I assume that is not the intention.

Correct. These are two different types of docs. Having the xmldoc-based docs though might mean that the conceptual docs can remain more conceptual.

> Classic Asserts, the Assert.That is also present there. I assume that is something about the filtering.

If there's a method called `That` on the asserts in the classic asserts project, it would show up there. 

____

Again thanks for the great feedback so far!

OsirisTerje:
> Classic Asserts, the Assert.That is also present there. I assume that is something about the filtering.

> If there's a method called That on the asserts in the classic asserts project, it would show up there.

It is not, but the classic asserts inherits the framework Assert class, which contains the That.  Can it be it picks that up?

```cs
 public abstract partial class ClassicAssert : Framework.Assert
```

OsirisTerje:
> You were not looking at the right place. 😆 At the top of the preview site is a new "API Reference" link -- that's what you'll want to click.

LOL !!!  My eyes was hiding that!

Never looked up there.  See that the menu also is there in the current docs page, but I have never noticed it !  

Perhaps we can do more out of that one ?   Ref #848

SeanKilleen:
@stevenaw 

> I understand the markdown docs are published as PRs are merged into this repo. Would framework-derived info like this be regenerating them as part of each release? How would versioning work over time as the API evolves across releases?

Great question! As of now, the way it works is the docs build points to a hard-coded release number of NUnit that I'd need to update each time. Our CI process pulls the asset from the release , unzips it, and moves it to a place where docfx  knows to look for it. Before this PR is merged I'll try to add a one-liner command in the dev container so people working with it there can easily pull in this part transparently.

Shortly after this is merged, I intend to follow up on it and set it to always pull the latest package from a semver perspective, which will keep it all up to date going forward.

SeanKilleen:
Alright, since the consensus is that people don't hate it, my next steps are:

* Get input on which stuff to exclude (feel free to drop lists of stuff here!) I'll try to ensure anything marked internal is auto-excluded.
* Get the local container tooling working so that the `build` helper command pulls the metadata as we expect and builds the whole experience. Probably with an additional one-liner command like `metadata`
* Ensure the README reflects this stuff
* Get it merged

SeanKilleen:
* The API docs are now automatically loaded into the right spot in the devcontainer
* The dev container now has an alias, `apidocs`, that will add or overwrite them again as needed, just in case it's needed.
* `build` and `serve` commands now also build the API docs.

PR #845, opened 2023-11-28
Minor wording adjustment to migration guide

stevenaw:
I've noticed a bit of hesitation on social media about the changes to assert styles. I thought a minor wording change could help here, mostly around dropping the word "convert" in one of the headings to minimize apparent friction.

I'm unsure which of these could be preferable so I've gone with the shorter, first alternative in this PR for now:

```diff
- #### Convert Classic Assert into NUnit 4.x equivalent
+ #### Using Classic Asserts in NUnit 4.x
```
or

```diff
- #### Convert Classic Assert into NUnit 4.x equivalent
+ #### Using Classic-style Asserts in NUnit 4.x
```

@manfred-brands @SeanKilleen I know you each put a lot of work into the guide, so I'd appreciate your thoughts on this

### DIFF
--- a/docs/articles/nunit/release-notes/Nunit4.0-MigrationGuide.md
+++ b/docs/articles/nunit/release-notes/Nunit4.0-MigrationGuide.md
@@ -96,7 +96,7 @@ Assert.That(actualText, Does.StartWith("42"), $"Expected '{actualText}' to start
 There are no code fixers for `FileAssert` and `DirectoryAssert`. They could be added, but we don't expect these to be
 used too much.
 
-#### Convert Classic Assert into NUnit 4.x equivalent
+#### Using Classic Asserts in NUnit 4.x
 
 If you want to keep the Classic Asserts and not convert them to the constraint model -- but do want to use the new NUnit
 4.x naming -- you'll need to update the code manually.
--- a/docs/articles/nunit/release-notes/Nunit4.0-MigrationGuide.md
+++ b/docs/articles/nunit/release-notes/Nunit4.0-MigrationGuide.md
@@ -96,7 +96,7 @@ Assert.That(actualText, Does.StartWith("42"), $"Expected '{actualText}' to start
 There are no code fixers for `FileAssert` and `DirectoryAssert`. They could be added, but we don't expect these to be
 used too much.
 
-#### Using Classic Asserts in NUnit 4.x
+#### Updating from Classic Asserts in NUnit 4.x
 
 If you want to keep the Classic Asserts and not convert them to the constraint model -- but do want to use the new NUnit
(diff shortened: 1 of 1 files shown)


docs/articles/nunit/release-notes/Nunit4.0-MigrationGuide.md
```diff
@@ -96,7 +96,7 @@ Assert.That(actualText, Does.StartWith("42"), $"Expected '{actualText}' to start
 There are no code fixers for `FileAssert` and `DirectoryAssert`. They could be added, but we don't expect these to be
 used too much.
 
-#### Convert Classic Assert into NUnit 4.x equivalent
+#### Using Classic Asserts in NUnit 4.x
```
manfred-brands, line 99: The steps outline how to migrate existing code to the new naming.
If you don't like `Convert`, as a minimum it should be `Update`.

stevenaw: That's a good idea, I like "Update" over "Convert" as it feels "lighter".

What are your thoughts on `Classic Asserts` vs `Classic-style Asserts` ? I'm trying to think of wording for this section which emphasizes to the reader that they can choose to stick with their chosen style of asserts if it is the classic style - even if it means they must update the referenced class.

SeanKilleen: I suggest `Updating from Classic Asserts in NUnit 4.x`

And I suggest it should be "Classic Asserts". We're calling the whole idea of them classic, not just the style.

OsirisTerje:
@stevenaw Can you link to some of these social media posts?   If there are many negative on this, we should rethink, or invite to a poll, to figure out what the overall response is.

stevenaw:
Agreed with @SeanKilleen re: moving of cheese

@OsirisTerje 
I wouldn't say "negative" so much as just a single comment on a twitter feed, but it had reminded me of @rprouse 's comment in the v4 planning issue about the friction of removing `ExpectedExceptionAttribute` in the move from v2 -> v3 and the desire he expressed at that point about making for an easy transition with whatever changes we included in v4. My knee-jerk reaction was then to think about softening language for that part of the docs.

No need in my mind either to conduct outreach at this time. I agree we've provided a clear migration path and considered several different scenarios in it for users.

stevenaw:
Thank you @SeanKilleen and @manfred-brands for your feedback. I've updated the section title to `Updating from Classic Asserts in NUnit 4.x`

manfred-brands (approved):
Thanks @stevenaw

SeanKilleen (approved):

PR #886, opened 2024-01-24
Updated running NUnitLite info

OsirisTerje:


docs/articles/nunit/running-tests/NUnitLite-Runner.md
```diff
@@ -16,22 +16,35 @@ With NUnit 3.0, the NUnitLite framework has been merged with the full NUnit fram
...
+ create a new console application.  If you use an existing test project you will get a warning about duplicate entry points. This can be ignored.
+
+Using a standard test project has the benefit that you use both the standard `dotnet test` command, the Visual Studio
+ Test Explorer and the NUnitLite runner to run your tests.
 
 * Make sure your test assembly references both **nunit.framework** and **nunitlite**.
 
-* The Main() for your console application should look something like this...
+```xml
+<PackageReference Include="NUnit" Version="4.0.1" />
```
SeanKilleen, line 28: ℹ️ Let me know if you're interested in turning this into a sample file so that we can keep the sample up to date with Dependabot and have the versions reflected here.

docs/articles/nunit/running-tests/NUnitLite-Runner.md
```diff
@@ -16,22 +16,35 @@ With NUnit 3.0, the NUnitLite framework has been merged with the full NUnit fram
...
-public static int Main(string[] args)
-{
-  return new AutoRun().Execute(args);
-}
+using NUnitLite;
+
+return new AutoRun().Execute(args);
 ```
 
+If you don't use top level statements, then just create a standard main routine with the same code.
```
SeanKilleen, line 40:
```suggestion
If you don't use top-level statements, then just create a standard main routine with the same code.
```

docs/articles/nunit/running-tests/NUnitLite-Runner.md
OsirisTerje: I would like that.  I added a whole sample test project outside now.  Was unsure how to do this with the existing snippets, because this is more than a snippet, it is a complete test project created first with `dotnet new nunit`, thne updated to the latest package versions, and then added the nunitlite stuff.

It would also be nice to replace the screenshot further down in the article, it is from 2014 and with and old version.  But the test should then have more than just a single Assert.Pass.

SeanKilleen:
Discussed with @OsirisTerje -- I'll:

* Add the NUnitLite project he created to snippets
* Reference the version pieces from the snippet
* Add some additional tests
* Update the screenshot

docs/articles/nunit/running-tests/NUnitLite-Runner.md
```diff
@@ -16,22 +16,31 @@ With NUnit 3.0, the NUnitLite framework has been merged with the full NUnit fram
...
 
-* The Main() for your console application should look something like this...
+* Make sure your test assembly references both **nunit.framework** and **nunitlite**.
 
-```csharp
-public static int Main(string[] args)
-{
-  return new AutoRun().Execute(args);
-}
+```xml
```
SeanKilleen, line 27: ℹ️ I ended up having to change this back because it looks like docfx has issues pulling in csproj files or xml files or something along those lines. Everything I tried came up blank.

Issue #902, opened 2024-02-19
Document TypeArgs for TestCase + TestCaseSource


stevenaw:
The changes for https://github.com/nunit/nunit/issues/1215 have been merged and should be documented

PR #903, opened 2024-02-19
Document TypeArgs for TestCase + TestCaseSource

stevenaw:
Fixes https://github.com/nunit/docs/issues/902

Build currently fails as it needs to update the version of NUnit to build the snippet


SeanKilleen:
@stevenaw sorry about the build issues. I don't think it's you; I think it's a permissions issue with how it's being run. It seems to be inconsistent but realizing there might be an access token that I misunderstood. Will investigate but let me know when this is ready and I'll check it out and merge it.

SeanKilleen:
@stevenaw I manually set the NUnit release version to pull into the docs. If it's in a later NUnit release, I may need to tweak that. I have a backlog item to make the build always pull the latest non-beta according to semver.

OsirisTerje (approved):
LGTM

OsirisTerje:
We now push all PRs to master as a package on myget.  Is it possible to use that on?

OsirisTerje:
Hmm, don't you see the package at all, or just in the feed ?

`> dotnet add package NUnit --version 4.1.0-alpha.0.3 --source https://www.myget.org/F/nunit/api/v3/index.json`

stevenaw:
Ok, snippets are now building! I think the remaining issue is related to having pushed to a branch on a fork

PR #927, opened 2024-04-16
Place explanations about filtering in their note block

chm-tm:
This is cosmetics.
I tried to meaningfully complete what seemed an unfinished first sentence.


OsirisTerje:
Thanks @chm-tm !  The "rules" here is a bit picky on linting :-)  Line length it looks like.

SeanKilleen:
This looks good to go -- the issue with the failing build in the last round of changes is just upon generating the deployment preview. I will maybe split that into a separate build step so that it's more obvious in the future, as it sometimes fails but doesn't indicate a PR problem necessarily.

Thank you for the contribution, @chm-tm!

PR #931, opened 2024-04-26
Close code block on NUnit2013 page

RenderMichael:
Currently broken: https://docs.nunit.org/articles/nunit-analyzers/NUnit2013.html#via-editorconfig-file


SeanKilleen (approved):
Thanks @RenderMichael! This looks good to me.

Pinging @mikkelbu because I think these docs are auto-generated so the true fix might be upstream.

PR #947, opened 2024-07-24
Document ThrowOnEachFailureUnderDebugger option

MaceWindu:
Fix https://github.com/nunit/nunit/issues/4765

@OsirisTerje , I've added adapter notes, but cannot find suitable document under `docs\articles\nunit\technical-notes`

### DIFF
--- a/docs/articles/nunit/technical-notes/usage/Debugging-Support.md
+++ b/docs/articles/nunit/technical-notes/usage/Debugging-Support.md
@@ -7,8 +7,6 @@ This document describes framework functionality for debugging support of tests a
 (From version 4.2.0)
 
 Setting name defined by `FrameworkPackageSettings.ThrowOnEachFailureUnderDebugger` constant.
-
-
 It may sometimes be necessary to debug the current state of a test during an `Assertion.Multiple` block. Typically,
 assertion exceptions are thrown only after the block is completed.
 
--- a/docs/articles/nunit/technical-notes/usage/Debugging-Support.md
+++ b/docs/articles/nunit/technical-notes/usage/Debugging-Support.md
@@ -9,7 +9,8 @@ This document describes framework functionality for debugging support of tests a
 Setting name defined by `FrameworkPackageSettings.ThrowOnEachFailureUnderDebugger` constant.
 
 
-It may sometimes be necessary to debug the current state of a test during an `Assertion.Multiple` block. Typically, assertion exceptions are thrown only after the block is completed.
+It may sometimes be necessary to debug the current state of a test during an `Assertion.Multiple` block. Typically,
+assertion exceptions are thrown only after the block is completed.
 
 This boolean framework setting will cause the debugger to throw on a failed assertion during an `Assert.Multiple`
 block. This allows user to break execution on exception and inspect tested code state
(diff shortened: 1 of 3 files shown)


OsirisTerje:
You can just add a new document under Technical Notes.  To be a bit future-proof, call it something like DebuggingSupport.md.  Then we can add more there later.

OsirisTerje:
I also see you have some lines that are longer than 120 chars.  You need to break those, unfortunately.

docs/articles/nunit/technical-notes/usage/Debugging-Support.md
```diff
@@ -0,0 +1,17 @@
...
+
+This document describes framework functionality for debugging support of tests and tested code.
+
+## `ThrowOnEachFailureUnderDebugger` Setting
+
+(From version 4.2.0)
+
+Setting name defined by `FrameworkPackageSettings.ThrowOnEachFailureUnderDebugger` constant.
+
+This boolean framework setting enables throwing of immediately catched assertion exception from `Assert.Multiple`
```
SeanKilleen, line 11:
```suggestion
This boolean framework setting enables throwing exceptions in `Assert.Multiple` immediately upon being caught.
```

docs/articles/nunit/technical-notes/usage/Debugging-Support.md
```diff
@@ -0,0 +1,17 @@
...
+block on each failed assertion. This allows user to break execution on exception and inspect tested code state
+at time of failure. In contrast default behavior of `Assert.Multiple` is to throw assertion exception only after
+whole block executed. In this situation tested code state could change or be unavailable.
```
SeanKilleen, line 14:
```suggestion

It may sometimes be necessary to debug the current state of a test during an `Assertion.Multiple` block. Typically, assertion exceptions are thrown only after the block is completed.

This boolean framework setting will cause the debugger to throw on a failed assertion during an `Assert.Multiple`
block. This allows user to break execution on exception and inspect tested code state
at time of failure.
```

docs/articles/vs-test-adapter/Tips-And-Tricks.md
```diff
@@ -378,6 +379,13 @@ If set, this setting will allow parallel execution of tests even if the debugger
 
 (From version 4.5.0)
 
+#### ThrowOnEachFailureUnderDebugger
+
+If set, this setting will allow handled assert exception for each failue in `Assert.Multiple` block if the debugger
+is attached. The default is `false`.
```
SeanKilleen, line 385:
```suggestion
If set, this setting will surface an assertion exception for each failure in an `Assert.Multiple` block if the debugger
is attached. The default is `false`.
```

SeanKilleen (commented):
Really nice contribution, @MaceWindu -- thank you!

I've made a few language suggestions for consideration. If those are merged in and they cause line size issues, I'll be happy to address those. 👍

MaceWindu:
@SeanKilleen thanks! Not a native speaker, so it is much appreciated

docs/articles/nunit/technical-notes/usage/Debugging-Support.md
```diff
@@ -0,0 +1,20 @@
...
+
+This document describes framework functionality for debugging support of tests and tested code.
+
+## `ThrowOnEachFailureUnderDebugger` Setting
+
+(From version 4.2.0)
+
+Setting name defined by `FrameworkPackageSettings.ThrowOnEachFailureUnderDebugger` constant.
+
+
```
SeanKilleen, line 11:
```suggestion

```

Removing extra blank line.

SeanKilleen (approved):
This LGTM 👍 It's possible the build will fail, but I checked and that's due to a long-standing issue in our CI process where our manual Netlify PR previews don't always work.

Issue #941, opened 2024-07-08
Use RenovateBot to update the API framework reference


SeanKilleen:
We can use a custom manager like I recently did in https://github.com/nunit/docfx-action/pull/57 to ensure there's a PR automatically created to update against the latest NUnit API docs.

SeanKilleen:
OK, on the last PR, I think I adapted it incorrectly. The regex for the fileMatch is written for a top-level file, which of course the build-process.yml file is not. So I'm going to do a PR to update the regex.

SeanKilleen:
Welp -- it would help if I used the right path syntax. Working in public is not at all embarrassing 😅

SeanKilleen:
There we go 🎉 I see the PR has just been created and is working as expected.

SeanKilleen:
I spoke too soon -- the PR is created and the text replacement is correct, but it's looking for a tag in NUnit of `v4.1.0` instead of `4.1.0`. -- I'll have to see if there's something else I can tweak. Will close this again if that ends up being doable within the suggestion PR itself.

SeanKilleen:
The build for #955 failed, but that's because the NUnit versions no longer appear to have tags starting with `v`, so I had to update that as well. Once I did, it passed exactly as I'd hoped.

So, going to once again close this. 👍

PR #952, opened 2024-07-28
Use RenovateBot for API docs version in build

SeanKilleen:
This introduces a RenovateBot custom manager.

In the `build-process.yml` file, it will look for the version in `NUNIT_VERSION_FOR_API_DOCS`, and will suggest an update for it when a new version of NUnit is published.

This ensures that our docs get PRs to stay up to date when a new version of NUnit is released.

Supports #941. Going to choose to leave it open for now because I have to merge this first before I can see if it's successful (I should see a PR generated to update to 4.1.0.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -12,7 +12,7 @@
       "matchStrings": [
         "NUNIT_VERSION_FOR_API_DOCS: \"(?<currentValue>.*?)\""
       ],
-      "depNameTemplate": "docfx",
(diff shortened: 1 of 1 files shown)


SeanKilleen:
Unfortunately there's no good way to test this (irony!) that I can see, and so I'm going to merge it and evaluate. I did check the regex in a regex tester and it worked as expected, plus I have the prior PRs I've done elsewhere with this technique.

PR #953, opened 2024-07-28
Update file match to be more precise

SeanKilleen:
Supports #941.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -7,7 +7,7 @@
     {
       "customType": "regex",
       "fileMatch": [
-        "^build-process.yml$"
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -7,7 +7,7 @@
     {
       "customType": "regex",
       "fileMatch": [
-        "^build-process.yml$"
+        "^.github\\workflows\\build-process.yml$"
```
SeanKilleen, line 10: ℹ️ I wasn't expecting the entire path to be necessary but it totally makes sense that it would be.

PR #959, opened 2024-08-05
Add documentation for EnterMultipleScope

### LINKED ISSUE #958: Document `Assert.EnterMultipleScope`
NUnit 4.2 adds `IDisposable EnterMultipleScope()` method in addition to the `Assert.Multiple()` delegate.

This issue is to update the documentation with this.

NUnit [PR 4587](https://github.com/nunit/nunit/issues/4587)

manfred-brands:
Fixes #958


OsirisTerje:
Suggest you add some comments on the difference between the old and new way  (and thus benefits of the new way) , including the disposables, and also point people to the fact there is no longer an action there.

Arhell (approved):

OsirisTerje:
@manfred-brands @SeanKilleen Could we manage to get this in ?  
I see there are some code issues here.

OsirisTerje:
Just fixed a few comments - due time to get it in, thanks for the writeup @manfred-brands !

PR #979, opened 2024-09-12
Adapter also supports NUnit 4

Bouke:
The name of the package is somewhat confusing now that NUnit 4 is out.


OsirisTerje (approved):

OsirisTerje:
Yes, we are aware of the name issue, but since there are no way to change or alias a package in nuget, we're stuck with the name.  
Thanks for finding that doc snag,  I thought we have them all covered, but....  more eyes are better !

SeanKilleen:
@Bouke just wanted to drop a quick line and say thank you for the contribution!

PR #997, opened 2024-12-20
Release nunit hotfix 4.3.1

OsirisTerje:


docs/articles/nunit/release-notes/framework.md
```diff
@@ -6,6 +6,63 @@ uid: frameworkreleasenotes
...
+
+### Acknowledgements
+
+We want to express our heartfelt gratitude to everyone who has contributed to this release
+by reporting bugs, suggesting enhancements, and providing valuable feedback.
+Your efforts help make NUnit better for the entire community.
+
+A special thank you to the following reporters for identifying issues:
+
+<table>
```
SeanKilleen, line 35: Doesn't have to hold up the PR at all, but I notice this is an HTML table rather than a markdown table. We should probably use markdown tables.

SeanKilleen (approved):
Approved with a minor suggestion and a comment.

docs/articles/nunit/release-notes/framework.md
OsirisTerje: I wanted it without header, since it is more of a 4 column list of names.  Didnt find any markdown way to do that.  But if have some tricks, please suggest :-)

SeanKilleen: I think it makes most sense as a bulleted list. Table probably shouldn't be used just to visually separate list items.

PR #998, opened 2024-12-21
Move table to valid markdown in release notes

SeanKilleen:
HTML is not valid in markdown and we need to avoid it where possible.

Similarly, using tables for purely visual layout is to be avoided, especially when there are not clear columns. It cuts things off on mobile and can be detrimental to the accessibility experience.


SeanKilleen:
@OsirisTerje I made the change the the table here, just FYI so that you don't have to. 👍

PR #1020, opened 2025-02-10
Auto-fix linting and remove line length linting

### LINKED ISSUE #887: GitHub Action update to do an auto-fix for markdown linting
the markdownlint-cli2 has the `-fix` flag to apply auto fixes where possible.

There's got to be a way to run this and then push the results back into the PR branch, to make it easier.

SeanKilleen:
After a quick conversation with @OsirisTerje we agreed that 1) it would be helpful to auto-fix markdownlint issues in docs on save, and 2) we didn't want the line length to cause issues in linting.

So, this PR:

* Adds the auto-fix on save functionality for Codespaces (and puts the same in the readme)
* Removes the line length linting
* Leaves the Rewrap plugin so that we can still have some friendly help on line length when desired
* Leaves `"editor.rulers": [120]` in settings so that we have a visual guideline.

Closes #887

### DIFF
--- a/.github/linters/.markdownlint.yml
+++ b/.github/linters/.markdownlint.yml
@@ -19,10 +19,7 @@
 # Rules by id #
 ###############
 MD024: false
-MD013: 
-  line_length: 120
-  code_blocks: false
-  tables: false
+MD013: false
 MD026: false
 MD046: false # Mix of indented and fenced code blocks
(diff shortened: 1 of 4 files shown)


SeanKilleen:
CC @OsirisTerje , this is in the main branch now. :)

PR #1018, opened 2025-02-10
Update MTP wording from Microsoft Test Platform to Microsoft.Testing.Platform

Youssef1313:

Issue #983, opened 2024-09-19
Docs build issue: Permissions when attempting to build a netlify preview


SeanKilleen:
When contributors send a PR, the build process fails on the `Start deployment (PR only)` step of our build process, similar to the below screenshot. The error message is `Input required and not supplied: ${name}`

This is often confusing for contributors and it would be good to get it resolved.

At this point in the build process, the PR has actually been built successfully and is ready to merge; it is only the deployment of the Netlify preview that is failing.

![image](https://github.com/user-attachments/assets/ab0debe3-455e-4773-befd-02e6aa2a9c0a)

SeanKilleen:
Happened to look into this and realized that it makes total sense.

Users outside of this repo are failing because we use a GitHub Actions secret, and those secrets are only available to repo collaborators:

> Anyone with collaborator access to this repository can use these secrets and variables for actions.

If can figure out a better way to share the token, this problem will go away.

SeanKilleen:
We are moving away from the Netlify build piece. As of #1063 this no longer exists. Therefore, closing this issue.

PR #1024, opened 2025-02-28
Update MTP wording in ToC

Youssef1313:
I previously updated the article in https://github.com/nunit/docs/pull/1018, but noticed now that I forgot to update it in ToC.


mikkelbu:
@SeanKilleen It seems the build is failing as we no longer supply a token in the "Start deployment (PR only)" step

SeanKilleen:
@mikkelbu this is a known issue: #983.

I'll review and approve.

PR #1039, opened 2025-05-04
Separate Netlify preview into its own build task

SeanKilleen:
Supports #983 by at least making it less confusing in terms of which aspects of the process are failing. This way folks can easily see that the build passed but the preview didn't.

Also the netlify preview task isn't required to pass in order to merge, so it should make that clearer too.


SeanKilleen:
image

PR #1048, opened 2025-06-09
chore: Make example compile

mikkelbu:


OsirisTerje:
Shouldn't we add this code to the snippet project and use that snippet in the markdown ?
Then it would also always be compiled there.

SeanKilleen:
> Shouldn't we add this code to the snippet project and use that snippet in the markdown ?
> Then it would also always be compiled there.

Agreed. I had a PR to do all of these but it got stale. Happy to do that here. 👍

PR #1062, opened 2025-08-31
Remove StringUtil from example DictionaryContainsKeyValuePairConstraint.

HeSitates:
In the docs StringUtil was still mentioned, but is removed.


OsirisTerje (approved):
We would like to have these code examples over to the snippets, so that they build all the time. That makes it easier to catch stuff like this.
But that would be a separate PR :-)

But thank you for your contribution @HeSitates !

Issue #172, opened 2017-02-27
Clarify use of IgnoreAttribute with TestCaseAttribute


CharliePoole:
@darthKwisatzHaderach commented on [Thu Sep 17 2015](https://github.com/nunit/nunit/issues/815)

Hi,

I have six test cases. I want to ignore two of them. When I do it as described below, I get error "Duplicate Ignore attribute.

```
[TestCase("03/03/20166"), Ignore("PRX-1000")]
[TestCase("03 03/2016"), Ignore("PRX-1000")]
[TestCase("0303/2016")]
[TestCase("033/03/2016")]
[TestCase("03/033/2016")]
[TestCase("03/03/201")]
public void InvalidDeadlineDateFormat(string dateFormat) {}
```

I am not exactly understand how to ignore these tests. Maybe it is a bug?

NUnit Version 3 3.0.0-beta-4

Best regards, Dmitrii

---

@rprouse commented on [Thu Sep 17 2015](https://github.com/nunit/nunit/issues/815#issuecomment-141094831)

Sounds like a bug to me :)

---

@darthKwisatzHaderach commented on [Thu Sep 17 2015](https://github.com/nunit/nunit/issues/815#issuecomment-141113190)

The same situation with NUnit 2.6.4.14350

---

@CharliePoole commented on [Thu Sep 17 2015](https://github.com/nunit/nunit/issues/815#issuecomment-141127973)

Actually, it's not a bug, but I'll warrant it's confusing. Here is what happens:

The textual layout of your attributes means nothing to .NET. When you use

``` C#
   [TestCase("03/03/20166"), Ignore("PRX-1000")]
   [TestCase("03 03/2016"), Ignore("PRX-1000")]
   [TestCase("0303/2016")]
```

it's just the same as

``` C#
   [TestCase("03/03/20166")]
   [Ignore("PRX-1000")]
   [TestCase("03 03/2016")]
   [Ignore("PRX-1000")]
   [TestCase("0303/2016")]
```

Or even...

``` C#
   [TestCase("03/03/20166")]
   [TestCase("03 03/2016")]
   [TestCase("0303/2016")]
   [Ignore("PRX-1000")]
   [Ignore("PRX-1000")]
```

The ignore attributes apply to the method and so would cause every test case to be ignored. But since you had two of them, you got a compiler error instead.

Since you are trying to ignore the individual test cases, you want this:

``` C#
   [TestCase("03/03/20166", Ignore="PRX-1000")]
[...]

PR #1069, opened 2025-09-10
Updated testcase documentation

OsirisTerje:
The issue https://github.com/nunit/nunit/issues/815  was documented for the Ignore attribute in #172, but it should also be updated for the TestCase, so this does that.

### DIFF
--- a/docs/articles/nunit/writing-tests/attributes/testcase.md
+++ b/docs/articles/nunit/writing-tests/attributes/testcase.md
@@ -49,7 +49,8 @@ TestCaseAttribute supports a number of additional named parameters:
 > [!WARNING]
 > When using the `Ignore` parameter (and others, see below), note that this has to be a named parameter. It is easy to accidentally add another `Ignore` attribute after the `TestCase` attribute. That will be the same as adding it separately, and it will apply to the complete fixture. This may apply to other named parameters, with names equal to other attributes, like the `Explicit` and `Category` parameters.
 
-Correct example usage
+Correct example usage:
+
 [!code-csharp[TestCaseWithIgnore](~/snippets/Snippets.NUnit/Attributes/TestCaseAttributeExamples.cs#TestCaseWithIgnore)]
 
 ![TestCaseIgnoreDoneCorrect](../../../../images/TestCaseIgnoreDoneCorrect.png)


docs/articles/nunit/writing-tests/attributes/testcase.md
```diff
@@ -42,6 +42,42 @@ TestCaseAttribute supports a number of additional named parameters:
   but may serve a purpose for the test author)
 * **TypeArgs** specifies the `Type`s to be used when targeting a generic test method. (_NUnit 4.1+_)
 
+## Be aware of mixing the syntax for named parameters and attributes with the same name
+
+### Ignore done right and wrong
```
SeanKilleen:
```suggestion
### The Correct `Ignore` Approach, by Example
```

docs/articles/nunit/writing-tests/attributes/testcase.md
```diff
@@ -42,6 +42,42 @@ TestCaseAttribute supports a number of additional named parameters:
...
+
+[!WARNING]
+When using the **Ignore** parameter (and others, see below), note that this has to be a named parameter. It is easy to add another Ignore attribute after the TestCase attribute. That will be the same as adding it separately, and it will apply to the complete fixture. Note, that this might apply to more of these named parameters, with names equal to other attributes, like the Explicit and Category parameters.
```
SeanKilleen:
```suggestion
When using the `Ignore` parameter (and others as noted below), keep in mind that this has to be a named parameter. It is easy to add another `Ignore` attribute after the `TestCase` attribute. That will be the same as adding it separately, and it would apply to the complete fixture. This also may apply to more of these named parameters, with names equal to other attributes, like the Explicit and Category parameters.
```

SeanKilleen:
```suggestion
> [!WARNING]
> When using the **Ignore** parameter (and others, see below), note that this has to be a named parameter. It is easy to add another Ignore attribute after the TestCase attribute. That will be the same as adding it separately, and it will apply to the complete fixture. Note, that this might apply to more of these named parameters, with names equal to other attributes, like the Explicit and Category parameters.
```

PR #1073, opened 2025-09-14
updated toc for constraints, as we missed the defaultconstraint, whic…

OsirisTerje:
…h is new


SeanKilleen:
Nice catch, thanks!

PR #1079, opened 2025-10-06
update zlib license link

Arhell:


SeanKilleen (commented):
I'm torn on this. On one hand, updating things seems good. On the other, my feeling is that the legacy docs are frozen in time and a little bit of link rot there may not be the worst thing. This PR isn't linked to a discussion.

@OsirisTerje , thoughts?

PR #1091, opened 2025-11-12
Updated adapter supported frameworks

OsirisTerje:


Arhell:
> @Arhell curious, why do you approve the PR when the build is failing?
> 
> If you are going to approve a PR, it would be really helpful to look at the failing build, see the errors, and suggest changes in a PR review.

Thank you, I understand, but I'm making the approved in advance because I'm sure that the errors will be corrected, and according to the document, everything seems to be fine.

OsirisTerje:
@Arhell I would appreciate commit suggestions when you see errors.  In this case it was edited on the web, so I missed those.  :-)

Arhell:
> @Arhell I would appreciate commit suggestions when you see errors. In this case it was edited on the web, so I missed those. :-)

Sure

PR #1092, opened 2025-11-13
fix snippets

OsirisTerje:
Fixed all warning in snippets project


SeanKilleen:
@OsirisTerje I'll fix this shortly. I think the issue is that I'm probably using latest rather than pinning to a version, and they likely added a new rule that we don't care about. Will investigate and fix ASAP.

SeanKilleen:
@mikkelbu yes, as I noted, it likely added a new rule that we need to disable.

I will fix this up.

SeanKilleen (approved):
👍 Approved and build fixed

PR #1173, opened 2026-05-03
fix: Make "Get latest NUnit Asset dir" work with `v` prefix

mikkelbu:
Releases are now prefixed with `v`


mikkelbu:
@OsirisTerje To fix the build after we started adding a `v` to the release tags

SeanKilleen (approved):

OsirisTerje (approved):
lgtm

Issue #1164, opened 2026-04-13
Document changes wrt to transfer to Action/Func


OsirisTerje:
See PR https://github.com/nunit/nunit/pull/5196 and discussion in issue https://github.com/nunit/nunit/issues/4824 

TestDelegate and ActionValueDelegate is changed to Action and Func.  See details in link to PR.

This is touching into a breaking change so we must document clearly what devs can experience and how they handle it.

PR #1183, opened 2026-05-29
Update Throws/Catch/DoesNotThrow async  assertion documentation for NUnit 5

### LINKED ISSUE #1180: NUnit 5 | Document that ThrowsAsync returns a Task
Version 5 of NUnit changes the signature of the following: 
- ThrowsAsync
- CatchAsync
- DoesNotThrowAsync

To return an awaitable. Documentation should be updated to reflect the new signature. The PR which introduces the change: https://github.com/nunit/nunit/pull/5268

Example pages: 
- https://docs.nunit.org/articles/nunit/writing-tests/assertions/classic-assertions/Assert.ThrowsAsync.html
- https://docs.nunit.org/articles/nunit/writing-tests/assertions/classic-assertions/Assert.CatchAsync.html
- https://docs.nunit.org/articles/nunit/writing-tests/assertions/classic-assertions/Assert.DoesNotThrowAsync.html

stevenaw:
Fixes #1180
Contributes to #1164

### DIFF
--- a/docs/snippets/Snippets.NUnit/Constraints/TypeConstraintSnippets.cs
+++ b/docs/snippets/Snippets.NUnit/Constraints/TypeConstraintSnippets.cs
@@ -25,7 +25,7 @@ public class TypeConstraintSnippets
 
         // Negative tests
         Assert.That("Hello", Is.Not.InstanceOf<int>());
-        Assert.That(null, Is.Not.InstanceOf<string>());        // null is not an instance of any type
+        Assert.That((string?)null, Is.Not.InstanceOf<string>());        // null is not an instance of any type
     }
     #endregion
 
--- a/docs/articles/nunit/writing-tests/assertions/classic-assertions/Assert.CatchAsync.md
+++ b/docs/articles/nunit/writing-tests/assertions/classic-assertions/Assert.CatchAsync.md
@@ -4,8 +4,26 @@ uid: classic-assert-catch-async
 
 # Assert.CatchAsync
 
-**Assert.CatchAsync** is similar to [Assert.ThrowsAsync](Assert.ThrowsAsync.md) but will pass for an exception
-that is derived from the one specified.
+**Assert.CatchAsync** is similar to [Assert.ThrowsAsync](Assert.ThrowsAsync.md) but will pass for an exception that is derived from the one specified.
+
+```csharp
+Task<Exception> Assert.CatchAsync(Func<Task> asyncCode);
+Task<Exception> Assert.CatchAsync(Func<Task> asyncCode,
+                            string message, params object[] params);
+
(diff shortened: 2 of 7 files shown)


docs/snippets/Snippets.NUnit/Constraints/TypeConstraintSnippets.cs
```diff
@@ -25,7 +25,7 @@ public void InstanceOfTypeConstraint_Examples()
 
         // Negative tests
         Assert.That("Hello", Is.Not.InstanceOf<int>());
-        Assert.That(null, Is.Not.InstanceOf<string>());        // null is not an instance of any type
+        Assert.That((string?)null, Is.Not.InstanceOf<string>());        // null is not an instance of any type
```
stevenaw, line 28: This resolved an overload ambiguity. The compiler seemed to want to resolve to the `Action` overload

stevenaw:
@OsirisTerje @SeanKilleen Can I ask your thoughts on merge timing of this one? It (currently) organizes these pages from an NUnit 5-first perspective, leaving previous behavior for a section later down the page. 

NUnit 5 should be pretty soon I think, but should we still wait to merge until after v5?

SeanKilleen:
@stevenaw I'm on a family vacation but will aim to provide comments tomorrow.

OsirisTerje:
@stevenaw 

I was thinking about the same.  Normally we just add a small note when we have added something, but for this major release there are whole major sections that needs a rewrite.  

Option 1 is to move all current documentation into a separate folder for Version 3 and 4 documentation. That would leave a lot of documentation unchanged, and all external links in to ours would be broken, as we don't use permalinks.  Alternatively move it to a separate web site. e.g. docs.v4.nunit.org.  External links would always point to the new pages, so it might be hard to find.  
Option 2 is to just move the V3 and 4 section down the page, as you mention, and mark that as for the V3/4, and do this only where applicable. This is the easy way out.

For Option two we could also add links at the top for going to V3/4, if the page is too large.

PS:  I like the way you written it in this PR.  Not sure if it scales for everything we have to change, like https://docs.nunit.org/articles/nunit/writing-tests/assertions/assertion-models/constraint.html  if we change the basic language as discussed in [#5244](https://github.com/nunit/nunit/issues/5244) and [#5221](https://github.com/nunit/nunit/issues/5221)

PS2: Perhaps DocFX can do something for us here, I read about Multi-Docset configurations.... @SeanKilleen ?

PS3: Or we could use the git versions, and mark the major version changes (tags), and publish one site per version, and only the current goes to the default doc site, the others go to subsites of that.  That could be doable.

OsirisTerje:
> NUnit 5 should be pretty soon I think, but should we still wait to merge until after v5?

Was hoping to get v5 out by the end of May..... oops,  too soon perhaps, but early June would be nice.
So once we agree on how to handle the versioning, just merge.

PR #1202, opened 2026-07-26
updated packages

OsirisTerje:
Renovate creates too many PRs, we need to fix that too


SeanKilleen:
> Renovate creates too many PRs, we need to fix that too

@OsirisTerje disagree -- it creates a separate PR per dependency, which is typically what you want.

However, there are ways to group dependencies if certain things always move together. We can look into that for sure.

OsirisTerje:
Yes, it's the grouping I'm thinking of.  Have that in nunit and adapter, and we have a template for it in .github repo.

